PR #59, opened 2024-01-11
feat(unleash): upgrade unleash to 5.6.10

bassrock:
## Goal

Upgrades unleash to 5.6.10 which is a prerequiste before going to 5.8.1.  

Also:
* Cleaned up the dependencies for feature flags, (didnt need the aws-sdk..)
* Fixed dev environment

## Implementation Decision

I had to turn off `skipLibCheck` until imports are fixed in unleash's type declaration files.


servers/feature-flags/tsconfig.json
```diff
@@ -2,7 +2,10 @@
   "extends": "tsconfig/graphql.json",
   "compilerOptions": {
     "outDir": "dist",
-    "rootDir": "src"
+    "rootDir": "src",
+    // Disable lib check because unleash has a ton of type errors in its declation files.
```
kschelonka, line 6: aha! this is what I ran into before

kschelonka (approved):

PR #81, opened 2024-01-18
fix(annotations-api): fixing flakey annotations tests

bassrock:
## Goal

Annotations api had some flakey tests that were inhibiting prs and causing us to re-run them 

https://mozilla-hub.atlassian.net/browse/POCKET-9477

### Batch Writes
This was causing 2 calls to the logAnnotationMutation that was hit intermittently based on what tests ran before  the batch tests. I got the test to always fail by adding users_meta to the seed list which caused a duplicate key error on every run. 

It seems that something about how a transaction runs it wasnt always deleting the old value first if logAnnotationMutation was called multiple times. I fixed this by only calling it once from the batch call, and making other calls conditional.

### system time

For some reason the createHighlight test had issues with Jest.fake timers unless I called setSystemTime.. no idea why. but in local testing this worked

### DIFF
--- a/servers/annotations-api/src/test/mutation/highlights-create.integration.ts
+++ b/servers/annotations-api/src/test/mutation/highlights-create.integration.ts
@@ -201,11 +201,8 @@ describe('Highlights creation', () => {
 
     it('should mark the list item as updated and log the highlight mutation', async () => {
       const updateDate = new Date(2022, 3, 3);
-
-      jest.useFakeTimers({
-        now: updateDate,
-        advanceTimers: true,
-      });
+      jest.useFakeTimers({advanceTimers: true});
+      jest.setSystemTime(updateDate);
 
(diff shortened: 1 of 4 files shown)


servers/annotations-api/src/test/mutation/highlights-create.integration.ts
```diff
@@ -201,11 +201,8 @@ describe('Highlights creation', () => {
 
     it('should mark the list item as updated and log the highlight mutation', async () => {
       const updateDate = new Date(2022, 3, 3);
-
```
bassrock, line 204: no idea why this is necessary to do it this way, but MULTIPLE runs and i didnt see the issue with this change.

kschelonka: wow, ok great find!!

kschelonka (approved):
🍠

PR #96, opened 2024-01-23
fix(terraform-modules): ensuring that the modules does not synth

bassrock:
## Goal

I noticed that the terraform modules was synth'ing during terraform plans/apply cause it had a synth script defined. This just renames it to avoid that conflict.


bassrock:
@kschelonka Gonna merge this cause I think its affecting builds/deploys.

PR #101, opened 2024-01-24
chore(deps): updating all dev dependencies

bassrock:
## Goal

Update all our current dev dependencies to be the latest versions.

Also made it so shared lists comes up locally without env-vars and doesnt keep trying to recreate the task def.


servers/shareable-lists-api/src/config/index.ts
```diff
@@ -20,8 +20,8 @@ export default {
     region: process.env.AWS_DEFAULT_REGION || 'us-east-1',
   },
   redis: {
-    primaryEndpoint: process.env.REDIS_PRIMARY_ENDPOINT || 'redis',
-    readerEndpoint: process.env.REDIS_READER_ENDPOINT || 'redis',
+    primaryEndpoint: process.env.REDIS_PRIMARY_ENDPOINT || 'localhost',
```
kschelonka, line 23: sweet

kschelonka (approved):

PR #107, opened 2024-01-25
fix(sha1): specifiying the git sha1 in the task definition

bassrock:
## Goal

I noticed that our task definitions were always set to the `latest` image. This would mean that if a scaling event occurred during a deployment or after a failed deploy, the tasks would not pull the correct currently "live" image. This change ensures that the task defintion always is pointing to a specific sha tag of the docker image.

### DIFF
--- a/packages/terraform-modules/src/base/ApplicationECSService.ts
+++ b/packages/terraform-modules/src/base/ApplicationECSService.ts
@@ -489,6 +489,7 @@ export class ApplicationECSService extends Construct {
       tags: this.config.tags,
       dependsOn: ecrRepos,
       provider: this.config.provider,
+      skipDestroy: true,
     });
 
     if (this.config.efsConfig) {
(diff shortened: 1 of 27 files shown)


packages/terraform-modules/src/base/ApplicationECSService.ts
```diff
@@ -485,6 +489,7 @@ export class ApplicationECSService extends Construct {
       tags: this.config.tags,
       dependsOn: ecrRepos,
       provider: this.config.provider,
+      skipDestroy: true,
```
bassrock, line 492: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_task_definition#skip_destroy

Makes it so that every PR doesn't say its destroying a resource if its just a deploy of new code.

kschelonka (approved):

PR #113, opened 2024-01-26
feat(list): enable users to update title of a save

kschelonka:
Mutation to update the title of a save. Just updates the user's personal save -- not the global metadata for the item stored by the parser.

Emits a new event for the updated title (analytics will need to integrate if they want to use)

### DIFF
--- a/servers/list-api/schema.graphql
+++ b/servers/list-api/schema.graphql
@@ -664,12 +664,15 @@ type Mutation {
 
   """
   Update the title display of a Saved Item, retrieved by ID.
-  This is user-save specific (does not update the metadata saved by the parser)
+  This is user-save specific (does not update the metadata saved by the parser).
+  Clients should ensure the input fits in the utf8mb3 character set (BMP only,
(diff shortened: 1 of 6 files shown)


servers/list-api/schema.graphql
```diff
@@ -665,6 +662,16 @@ type Mutation {
...
   updateSavedItemUnFavorite(id: ID!): SavedItem!
 
+  """
+  Update the title display of a Saved Item, retrieved by ID.
+  This is user-save specific (does not update the metadata saved by the parser)
+  """
+  updateSavedItemTitle(
+    id: ID!
+    timestamp: ISOString!
+    title: String! @constraint(minLength: 1, maxLength: 80)
```
bassrock, line 672: Whats the DB limit?

bassrock (approved):
LGTM

servers/list-api/schema.graphql
kschelonka: checking

PR #115, opened 2024-01-28
fix(plugins): ensure standard apollo and sentry usage

bassrock:
## Goal

Our sentry and apollo usage should be the same within all our microservices. 

This pulls most of the setup all into the apollo package and then ensures that all servers utilize it

Some Changes
* Always enable the GraphQL Landing page, since all subgraphs are in a VPC
* Include errors to apollo studio
* Enable inline tracing for the router
* Disable the usage reporting on individual subgraphs.
* Let sentry auto discover plugins to use
* Switch Sentry to the encoded id and context in our apollo sentry request handler


.vscode/settings.json
```diff
@@ -3,5 +3,46 @@
     {
       "mode": "auto"
     }
-  ]
+  ],
+  "hidefiles.configurationType": "local",
```
kschelonka, line 7: cool

packages/apollo-utils/src/sentry/apolloSentryPlugin.ts
```diff
@@ -61,7 +61,10 @@ export const sentryPlugin: ApolloServerPlugin<BaseContext> = {
           const operationVariablesJson = JSON.stringify(ctx.request.variables);
           const requestId = ctx.request.http?.headers.get('x-graph-request-id');
           const requestTraceId = ctx.request.http?.headers.get('x-amzn-trace');
-          const requestUserId = ctx.request.http?.headers.get('userid');
+          const requestEncodedId = ctx.request.http?.headers.get('encodedId');
```
kschelonka, line 64: 👍

packages/apollo-utils/src/sentry/apolloSentryPlugin.ts
```diff
@@ -81,32 +83,38 @@ export const sentryPlugin: ApolloServerPlugin<BaseContext> = {
...
 
-            // report error
-            Sentry.captureException(err);
+          if (apiId !== undefined) {
+            scope.setTag('pocket-api-id', apiId);
+          }
+
+          scope.setUser({
+            id: (requestEncodedId as string) || undefined,
+            ip_address: (gatewayIpAddress as string) || undefined,
```
kschelonka, line 113: we added the x-forwarded-for final value as a header, I think called originIp? as an alternative

kschelonka (approved):
lgtm, optional about adding the other ip address source to the observability layer

packages/apollo-utils/src/sentry/apolloSentryPlugin.ts
bassrock: @kschelonka is that in the router? I was using the docs on https://studio.apollographql.com/graph/pocket-client-api/variant/current/home

PR #116, opened 2024-01-29
fix: swallow CorpusItem.timeToRead resolver errors

jpetto:
- these errors would be from the parser, which we do not want impacting any clients. this field is nullable on CorpusItem, so returning null on error works


bassrock (approved):

efixler (approved):
lgtm!

PR #131, opened 2024-01-31
chore(release): removing pushing back a changelog

bassrock:
## Goal

Pushing a changelog required unrestricted push access to the monorepo, which is too risky to publish a module.


packages/terraform-modules/package.json
```diff
@@ -30,13 +30,7 @@
     "parse-domain": "5.0.0"
   },
   "devDependencies": {
-    "@semantic-release/changelog": "6.0.3",
-    "@semantic-release/commit-analyzer": "11.1.0",
-    "@semantic-release/exec": "6.0.3",
-    "@semantic-release/git": "10.0.1",
```
bassrock, line 36: Most of these are all included by default in sem release, so no need to also manage their versions here

kschelonka (approved):

PR #142, opened 2024-02-01
feat(shareable-lists): paginated Items on a List

kschelonka:
Provides pagination API for items on a list, deprecating the old API which returned all Items in the list in a single response.

This is fine for the majority of cases, but we already see users with thousands of items in their lists _even when they have to do it one-at-a-time_. Once the bulk api is enabled in the frontend, we can probably anticipate that list sizes will grow.

**Changes**
* Update the pagination validation util -- make the type responses more specific/accurate, and optionally disable the cursor validator
* Add a paginated field `items` to ShareableListItem, which implements the Relay cursor pagination spec; deprecate `listItems` field
* Fake cursor pagination in the backend (decode into offset pagination)

**Not Changed**
* the Admin API - my reasoning is we're not allowing public lists to be created, and we'll update the API if/when we need to have administrative access


servers/shareable-lists-api/src/database/mutations/ShareableList.ts
```diff
@@ -173,6 +173,7 @@ export async function createAndAddToShareableList(
   const completeList: ShareableList = {
     ...list,
     // bigint causes issues again...
+    id: list.id as unknown as bigint,
```
bassrock, line 176: 🤣

servers/shareable-lists-api/src/models/ShareableListItem.ts
```diff
@@ -0,0 +1,48 @@
...
+  constructor(private context: BaseContext) {}
+  byListId(
+    listId: number,
+  ): SelectQueryBuilder<DB, 'ListItem', ListItemResponse> {
+    const qb = this.context.conn
+      .selectFrom('ListItem')
+      .select(ListItemSelect)
+      .orderBy('sortOrder', 'asc')
+      .orderBy('createdAt', 'asc')
+      // Temporary tiebreaker sort for deterministic response
```
bassrock, line 18: whats the long term?

servers/shareable-lists-api/src/database/OffsetPagination.ts
```diff
@@ -0,0 +1,126 @@
...
+ * have a consistent experience with our cursor-based
+ * pagination pattern (even when offset is more appropriate)
+ */
+export class OffsetPagination<TB extends keyof DB, O> {
+  public static sentinel = '__pkt__';
+  /**
+   * Create a fake cursor from an offset value
+   */
+  public static encodeCursor(n: number): string {
+    // Probably unnecessary, but give the "cursors" a little oomph
```
bassrock, line 18: 💥

bassrock (approved):
lgtm! really nice work on the pagination here

servers/shareable-lists-api/src/models/ShareableListItem.ts
kschelonka: either changing the sortOrder field so it doesn't allow collisions if sortOrder isn't explicitly set, or staying with this (or any other numeric ID)

servers/shareable-lists-api/src/database/mutations/ShareableList.ts
kschelonka: I know... should I be better? yes. but is a number in javascript going to exceed bigint realistically?

PR #145, opened 2024-02-02
fix(cache): clear dataloader cache on refresh

bassrock:
## Goal

I played with adding a refresh call to web-client and noticed that the refresh mutation was not clearing the dataloader cache and so that was returning stale data still.


servers/parser-graphql-wrapper/src/resolvers.ts
```diff
@@ -205,6 +205,15 @@ export const resolvers = {
         maxAge: info.cacheControl.cacheHint.maxAge,
       });
       const item = await getItemByUrl(url);
+
+      // Clear our dataloader cache
+      clear({
```
kschelonka, line 210: 👍

kschelonka (approved):

PR #155, opened 2024-02-06
fix(pocket-graph): do not forward anonymous userid

kschelonka:
Do not forward 'anonymous' as userId if that's the sub claim in the JWT.

Add additional logic in subgraphs which check for the userId header to be extra safe, although it should be unnecessary.

This should resolve [POCKET-9492]. The error occurs when list-api mutations fail due to attempting to write an invalid value of userId to the `list` table.

[POCKET-9492]: https://mozilla-hub.atlassian.net/browse/POCKET-9492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


servers/annotations-api/src/context.ts
```diff
@@ -53,13 +53,12 @@ export class ContextManager implements IContext {
   get userId(): string {
     const userId = this.config.request.headers.userid;
 
-    if (!userId) {
+    if (!userId || userId === 'anonymous') {
```
bassrock, line 56: if we are removing it at the router do we need this?

bassrock (approved):

PR #165, opened 2024-02-09
feat(parser): resolve item from short url

kschelonka:
Adds ability to resolve an Item entity from a short Pocket Share URL. The Pocket Share URL is created on request and redirects to the original page (currently controlled by Web repo).

[POCKET-9547]

[POCKET-9547]: https://mozilla-hub.atlassian.net/browse/POCKET-9547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/servers/parser-graphql-wrapper/src/resolvers.ts
+++ b/servers/parser-graphql-wrapper/src/resolvers.ts
@@ -5,7 +5,11 @@ import config from './config';
 import { MediaTypeParam, ParserAPI } from './datasources/parserApi';
 import { MarticleElement, parseArticle } from './marticle/marticleParser';
 import { CacheScope } from '@apollo/cache-control-types';
-import { getShortUrl } from './shortUrl/shortUrl';
+import {
+  getShortUrl,
+  extractCodeFromShortUrl,
+  itemIdFromShareCode,
+} from './shortUrl/shortUrl';
 import { IContext } from './context';
 import { generateSSML } from './ssml/ssml';
 import { serverLogger } from '@pocket-tools/ts-logger';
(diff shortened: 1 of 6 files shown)


servers/parser-graphql-wrapper/src/resolvers.ts
```diff
@@ -139,8 +159,19 @@ export const resolvers = {
     ): Promise<Item> => {
       return getItemById(id, await repositories.itemResolver);
     },
-    itemByUrl: async (_source, { url }): Promise<Item> => {
-      return getItemByUrl(url);
+    itemByUrl: async (_source, { url }, { repositories }): Promise<Item> => {
+      // If it's a special short share URL, use alternative resolution path
+      const shortCode = extractCodeFromShortUrl(url);
```
bassrock, line 164: oooo i like how you did this

servers/parser-graphql-wrapper/src/resolvers.ts
```diff
@@ -111,6 +115,11 @@ export const resolvers = {
     },
     shortUrl: async (parent, args, context: IContext): Promise<string> => {
       const repo = await context.repositories.sharedUrlsResolver;
+      // If the givenUrl is already a short share url, return the same value
+      // to avoid another db trip
+      if (extractCodeFromShortUrl(parent.givenUrl) != null) {
```
bassrock, line 120: 🧠

bassrock (approved):
I really like how you implemented this in the existing itemByURL, that way all of this is transparent to the client and we don't need a new mutation. Thats amazing.

PR #167, opened 2024-02-12
fix(redis): enabling cluster support for serverless

bassrock:
## Goal

Get serverless redis working by supporting cluster mode in the client which is used by AWS Serverless Redis.

In a followup, need to ensure that if redis is down, or we can't connect that it fails silently to the user.

Followups
* https://mozilla-hub.atlassian.net/browse/POCKET-9599
* https://mozilla-hub.atlassian.net/browse/POCKET-9600

### DIFF
--- a/servers/image-api/src/server/apollo.ts
+++ b/servers/image-api/src/server/apollo.ts
@@ -17,6 +17,7 @@ import {
 } from '@pocket-tools/apollo-utils';
 import { setMorgan, serverLogger } from '@pocket-tools/ts-logger';
 import config from '../config';
+import { getRedis } from '../cache';
 
 /**
  * Context factory function. Creates a new context upon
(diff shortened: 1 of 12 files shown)


servers/image-api/src/server/apollo.ts
```diff
@@ -60,7 +61,16 @@ export async function startServer(port: number): Promise<{
   app.use(Sentry.Handlers.tracingHandler());
 
   // Expose health check url
-  app.get('/.well-known/apollo/server-health', (req, res) => {
+  app.get('/.well-known/apollo/server-health', async (req, res) => {
```
bassrock, line 64: Ensure that we can connect to redis before declaring the task healthy.

kschelonka (approved):

PR #182, opened 2024-02-20
feat(list): offset-pagination for v3 api proxy

kschelonka:
Adds an internal API for offset pagination in order to support v3 api proxy (which uses offset pagination scheme).

[POCKET-9613]

### DIFF
--- a/servers/list-api/schema.graphql
+++ b/servers/list-api/schema.graphql
@@ -118,6 +118,16 @@ input PaginationInput {
   last: Int
 }
 
+"""
+Input for offset-pagination (internal backend use only).
+"""
+input OffsetPaginationInput {
+  """ Defaults to 0 """
+  offset: Int @constraint(min: 0)
+  """ Defaults to 30 """
+  limit: Int @constraint(exclusiveMin: 0, max: 5000)
+}
+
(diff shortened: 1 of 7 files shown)


servers/list-api/schema.graphql
```diff
@@ -546,6 +556,16 @@ type SavedItemTagAssociation {
   tagId: ID!
 }
 
+"""
+A page of SavedItems, retrieved by offset-based pagination.
+"""
+type SavedItemsPage @tag(name: "internal") {
```
bassrock, line 562: i like starting to tag these

servers/list-api/schema.graphql
```diff
@@ -118,6 +118,16 @@ input PaginationInput {
...
 }
 
+"""
+Input for offset-pagination (internal backend use only).
+"""
+input OffsetPaginationInput {
+  """ Defaults to 0 """
+  offset: Int @constraint(min: 0)
+  """ Defaults to 30 """
+  limit: Int @constraint(exclusiveMin: 0, max: 5000)
```
bassrock, line 128: Is 5000 what existing api allows!?  

Does list build its schema already?

bassrock (approved):
LGTM

PR #193, opened 2024-02-21
fix(apollo-utils): ensuring that apollo utils does not contain workspace packages

bassrock:
# Goal

I noticed that https://github.com/Pocket/content-monorepo/pull/66 was failing because ts-logger wasn't changed to a published version when released with semantic release.  While there are ways around this, I evaluated it's usage in apollo-utils and came to a conclusion that we probably don't need it there and can side step this publishing issue for now.


bassrock:
Closing and will find a way to fix the deployments when its needed or asked for.

PR #208, opened 2024-02-23
feat(v3proxy): add complete and simple detailType queries (/get)

kschelonka:
Ripping the band-aid off for the most involved change.

### Changes

* Remove the cursor-based pagination queries
* Split off execution of the /get request depending on detailType query param ('simple'/'complete')
    * if missing or incorrect, default to 'simple'
* Add 'complete' query
    * Updated graphql fragments
    * Field transformers
    * Test data 
* Fix logic for some conditional fields based on actual v3 data (e.g. top_image_url key is missing if there is no data, rather than an empty string)
* Add the sortId 

### Didn't Do

* listen length estimate (added ticket as this field was not available)
* adding the undocumented top-level fields (turns out "complete" returns 1 for both detailType=simple and detailType=complete? so that's not what it means...)
* change old test fixtures -- I fixed them, but they really need to be made explicit to avoid bugs in the future


servers/v3-proxy-api/src/graph/toRest.ts
```diff
@@ -3,17 +3,263 @@
...
+  VideosItemObject,
+  VideoTypeMap,
   ListItemObject,
-  RestResponse,
+  ListItemObjectComplete,
+  ListItemObjectAdditional,
 } from './types';
+import { RestResponseSimple, RestResponseComplete } from './types';
+
+/**
```
kschelonka, line 30: My original version of these reducers was getting too clever with factory functions and generics... Describing them explicitly like this is verbose but I think way more readable.

servers/v3-proxy-api/src/graph/toRest.ts
```diff
@@ -3,18 +3,265 @@
...
+  }
+  const completeFieldMap = {
+    authors: AuthorsReducer(savedItem.item.authors, savedItem.id),
+    domain_metadata: DomainMetadataTransformer(savedItem.item.domainMetadata),
+    images: ImagesReducer(savedItem.item.images, savedItem.id),
+    tags: TagsReducer(savedItem.tags, savedItem.id),
+    videos: VideosReducer(savedItem.item.videos, savedItem.id),
+    image: DisplayImageTransformer(savedItem.item.images, savedItem.id),
+  };
+  const complete = Object.entries(completeFieldMap).reduce(
```
bassrock, line 185: 👏

servers/v3-proxy-api/src/graph/toRest.ts
```diff
@@ -3,18 +3,265 @@
...
+): ListItemObject | ListItemObjectComplete {
+  const baseFields = {
+    item_id: savedItem.id,
+    favorite: savedItem.isFavorite ? ('1' as const) : ('0' as const),
+    status: savedItem.isArchived ? ('1' as const) : ('0' as const),
+    time_added: savedItem._createdAt?.toString(),
+    time_updated: savedItem._updatedAt?.toString(),
+    time_read: (savedItem.archivedAt ?? '0').toString(),
+    time_favorited: (savedItem.favoritedAt ?? '0').toString(),
+    // TODO POCKET-9657
```
bassrock, line 214: We need to add in the listen api in general to this repo and the graph..

bassrock (approved):
lgtm!!

PR #212, opened 2024-02-27
feat(user-list-search-indexing): break out indexing lambdas logic

bassrock:
## Goal

Modify lambdas to deploy outside of the main user list search server code 

https://mozilla-hub.atlassian.net/browse/POCKET-9665

#implementation decisions

I got a bit carried away and switched all the code to use Knex so that we had a singular pattern for how we operate, by doing that and extracting the DB logic from the lambdas I was able to cleanup the entire mysql initilization in this service.


docker-compose.yml
```diff
@@ -36,7 +36,7 @@ services:
       retries: 15
 
   redis:
-    image: redis:7.1 # match aws elasticache
+    image: redis:7 # we use 7.1 on aws, but redis pulled that tag....
```
kschelonka, line 39: 😬 good catch I was wondering why I was having trouble building

lambdas/user-list-search-indexing/src/helper.ts
```diff
@@ -0,0 +1,86 @@
+import { UserItemsSqsMessage, UserListImportSqsMessage } from './types';
+import { nanoid } from 'nanoid';
+import fetch from 'node-fetch';
+import { config } from './config';
+
+export const processUserImport = async (
```
kschelonka, line 6: Mind slapping some short docstrings on these guys to help future readers?

lambdas/user-list-search-indexing/src/types.ts
```diff
@@ -0,0 +1,16 @@
+//Dupe of types from user-list-search-sq-to-kinesis.. may be worthwile to make this a package..
```
kschelonka, line 1: 🤔

servers/user-list-search/src/server/routes/itemDelete.ts
```diff
@@ -0,0 +1,91 @@
...
+      itemIds,
+    });
+
+    const listItems = itemIds.map((itemId) => {
+      return { itemId, userId };
+    });
+
+    if (listItems.length > 0) {
+      try {
+        await deleteFromElasticSearch(listItems);
```
kschelonka, line 76: Does this finish pretty quickly?

bassrock: so/so. I had this as async on the node background but then the tests were flakey. Which got me thinking.. it probably should not take longer then the lambda execution timeout, and we don't want lambda to just go crazy hitting this api, we should probably wait for it to finish. It also makes the lambda responsible for retries instead of it just failing.. 

i dont hold a strong opinion on this one given the current environmet.

docker-compose.yml
bassrock: so... annoying.. why would you remove a published tag!!!

lambdas/user-list-search-indexing/src/helper.ts
bassrock: done!

PR #214, opened 2024-02-27
feat(schema): [MC-573] Extend CorpusItem with `datePublished` field

nina-py:
## Goal

Extend CorpusItem with `datePublished` field. Added this to be able to show publication date for curated corpus items on the Schedule page in Curation Admin Tools.

- Added a resolver, integration tests.

- Updated .gitignore to ignore WebStorm config files.

## Reference 

https://mozilla-hub.atlassian.net/browse/MC-573

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ coverage
 # misc
 .DS_Store
 *.pem
+.idea
 
 # debug
 npm-debug.log*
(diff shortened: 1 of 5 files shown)


nina-py:
Turns out we'll be going the way of storing this field in the Curated Corpus database instead. Closing this as not needed.

PR #236, opened 2024-03-01
fix(images): ensuring that Images always have a URL response field

bassrock:
## Goal

Ensure images always have a URL response field and not null

https://mozilla-hub.atlassian.net/browse/POCKET-9681

Basically any client that was requesting URL on image... was getting an internal server error....


kschelonka (approved):
Follow-up to add e2e graph test coverage which would catch this

PR #244, opened 2024-03-04
fix: Snowplow retries on network error

mmiermans:
# Goal

Work around a bug in Snowplow, where it does not retry on error if method [is set to Post](https://github.com/Pocket/pocket-monorepo/blob/main/servers/shared-snowplow-consumer/src/snowplow/tracker.ts#L13).

# Todo

- [ ] Add test coverage using MSW, similar to [this](https://github.com/Pocket/content-monorepo/blob/d212be501fbd3a038c4b5a4131eb2b22dc3299d6/packages/content-common/events/snowplow/index.spec.ts)?

# References

- Issue filed with Snowplow: https://github.com/snowplow/snowplow-javascript-tracker/issues/1290


kschelonka:
Closing due to inactivity

PR #253, opened 2024-03-05
fix(parser): fix short code generation for non-https

kschelonka:
Fix incorrect test assertion and ensure http protocol is used for non-secure short code urls (https for secure).


bassrock (approved):

kschelonka:
This actually isn't a bug? Or if it is, the web repo does the same thing...

PR #262, opened 2024-03-07
feat(search): add offset-based searches to support proxy

kschelonka:
Some code reuse, some copying where it would be more annoying to refactor for reuse. Hopefully we'll be able to deprecate searchSavedItems in favor of consolidating on the advanced search, and remove the deprecated search api as well.

[POCKET-9630]

### DIFF
--- a/servers/user-list-search/src/datasource/elasticsearch/elasticsearchSearch.ts
+++ b/servers/user-list-search/src/datasource/elasticsearch/elasticsearchSearch.ts
@@ -1,4 +1,4 @@
-import { GetResponse } from 'elasticsearch';
+import { GetResponse, SearchResponse } from 'elasticsearch';
 import { client } from './index';
 import { config } from '../../config';
 import {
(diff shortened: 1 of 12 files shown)


servers/user-list-search/src/datasource/elasticsearch/elasticsearchSearch.ts
```diff
@@ -597,9 +604,68 @@ export async function advancedSearch(
...
+    offset: params.pagination?.offset ?? 0,
+  };
+}
+
+export async function searchSavedItemsByOffset(
+  params: SearchSavedItemOffsetParams,
+  userId: string,
+): Promise<SavedItemSearchResultPage> {
+  const { result, searchParams } = await searchBase(params, userId);
+  const entries: SavedItemSearchResult[] = result.hits.hits.map((res: any) => ({
```
bassrock, line 636: Supoosed to be `hits.hits`?

bassrock (approved):

servers/user-list-search/src/datasource/elasticsearch/elasticsearchSearch.ts
kschelonka: yep, if you can believe it...

PR #266, opened 2024-03-08
feat(v3proxy): add search query to proxy

kschelonka:
Adds search functionality to the proxy.

** Changes **
* Call the search api if the `search` parameter is populated
* Supports the `total`, and `detailType` input parameters changing the response shape.
* Updates behavior for default sort when it is not provided -- if "search" is populated, default sort will be relevance rather than newest so that we can provide the best sort order available (in the graph API, if the user is not premium, it will be reset to newest). Non-search queries still default to "newest".

** Other Changes **

* fix a typo in the v3 response type - `cachetype`, not `cacheType`
* the sort/filter classes are just functions now. I had some plans to reuse the classes and generate the sort objects for each query, but due to the different types generated by the grahpql library it just became too complicated. Repetitive functions is the name of the game for this proxy anyway, so...


servers/v3-proxy-api/src/graph/get/toGraphQL.spec.ts
```diff
@@ -1,132 +1,246 @@
...
-            status: 'ARCHIVED',
-            contentType: 'IS_READABLE',
-          },
-        },
-      },
-    ])(
-      'maps filters appropriately, including combos',
-      ({ params, expected }) => {
-        const actual = setSaveInputsFromGetCall({ ...defaultQuery, ...params });
+  describe('get saved items', () => {
```
kschelonka, line 11: Nesting this test muddied up the diff, sorry

servers/v3-proxy-api/src/graph/get/toGraphQL.ts
```diff
@@ -101,33 +129,89 @@ export class Filter {
...
-  }
+    },
+    {} as SearchFilterInput,
+  );
+  return Object.freeze(filter);
 }
 
-export function setSaveInputsFromGetCall<T extends V3GetParams>(
-  requestParams: T,
+export function setSavedItemsVariables(
```
kschelonka, line 193: This diff is weird because of the repetitive lines

bassrock (approved):

PR #273, opened 2024-03-11
fix(v3proxy): fix filters and total+complete response

kschelonka:
Do not pass an empty filters object (graphql is rejecting it), instead only add filters to the variables if they are relevant.

Fix a bug where the Simple transformer was being called when it should have been the Complete transformer.

### DIFF
--- a/servers/v3-proxy-api/src/graph/get/toGraphQL.ts
+++ b/servers/v3-proxy-api/src/graph/get/toGraphQL.ts
@@ -195,14 +195,18 @@ export function setSavedItemsVariables(
 ): UserSavedItemsByOffsetArgs {
   const filter = SavedItemsFilterFactory(requestParams);
   const sort = SavedItemsSortFactory(requestParams);
-  return {
+  const variables = {
     pagination: {
       limit: requestParams.count,
       offset: requestParams.offset,
     },
     sort,
-    filter,
   };
+  // Attach filter only if not empty
+  if (Object.keys(filter).length) {
+    variables['filter'] = filter;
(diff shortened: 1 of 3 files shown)


servers/v3-proxy-api/src/routes/v3Get.ts
```diff
@@ -120,7 +121,7 @@ export async function processV3call(
         variables,
       );
       if (data.total) {
-        return savedItemsSimpleTotalToRest(response);
+        return savedItemsCompleteTotalToRest(response);
```
bassrock, line 124: 😅

servers/v3-proxy-api/src/graph/get/toGraphQL.ts
```diff
@@ -195,28 +195,37 @@ export function setSavedItemsVariables(
...
     term: requestParams.search,
     pagination: {
       limit: requestParams.count,
       offset: requestParams.offset,
     },
     sort,
-    filter,
   };
+  // Attach filter only if not empty
+  if (Object.keys(filter).length) {
```
bassrock, line 227: i kinda wish graphql would be ok with empty objects so clients didnt have to do this kind of thing.

bassrock (approved):

servers/v3-proxy-api/src/graph/get/toGraphQL.ts
kschelonka: I thought it would since all the params are optional, but it's rejecting it :(

PR #276, opened 2024-03-12
fix(v3proxy): return user-provided title as given_title if exists

kschelonka:
Request SavedItem.title in the response and return it as given_title if not null. Otherwise fall back to resolved title (SavedItem.item.title).

Pending items will only return a non-empty string if the title is provided by the user (there is no resolved title to fall back on).

[POCKET-9724]

[POCKET-9724]: https://mozilla-hub.atlassian.net/browse/POCKET-9724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bassrock (approved):
short and sweet

PR #286, opened 2024-03-13
fix(since): ensuring that since is converted to a number

bassrock:
## Goal

Make sure that since is always converted to an integer


servers/v3-proxy-api/src/routes/validations/GetSchema.ts
```diff
@@ -108,6 +108,9 @@ export const V3GetSchema: Schema = {
         min: 0,
       },
     },
+    customSanitizer: {
+      options: (value) => (typeof value === 'string' ? parseInt(value) : value),
+    },
```
kschelonka, line 113: This works but I think you can just use the builtin toInt

```suggestion
    toInt: true,
```

kschelonka (approved):

PR #289, opened 2024-03-14
feat(int-mask): moving int-mask to its own package

bassrock:
## Goal

Move int mask to it's own package so we can utilize it for things like itemIds.


packages/int-mask/src/mask.ts
```diff
@@ -1,50 +1,63 @@
...
 
 //mimicing web repo behavior:
 //https://github.com/Pocket/Web/blob/881d5b51646b3ccd0d7f7c7206510f263d9fef03/shared/classes/IntMask.php
 // LETTER_INDEX is used in combination with MD5_RANDOMIZER to map the few
 // letters returned in an md5, [a-f], into a larger set of characters to
 // contribute the the obfuscation, and is less clearly a md5.
-const LETTER_INDEX = new Map<string, number>(config.secrets.letterIndex);
-
-const MD5_RANDOMIZER = new Map<string, string[]>(config.secrets.md5Randomizer);
+type LetterRandomizerOptions = {
```
kschelonka, line 9: yes

servers/user-api/src/config/index.ts
```diff
@@ -87,114 +87,116 @@ export default {
...
-          ['b', 2],
-          ['c', 8],
-          ['d', 9],
-          ['e', 1],
-          ['f', 0],
-          ['0', 9],
-        ],
-    salt1: process.env.SALT_1 || '123asdf',
-    salt2: process.env.SALT_2 || 'asdaa47',
+    intMask: {
```
kschelonka, line 90: oh indentation

kschelonka (approved):
🦅

PR #290, opened 2024-03-14
feat(itemmask): adding a masked item id

bassrock:
## Goal

Allow generation of an ID for future use in getpocket.com/read/someitem so that links can be shareable.


servers/parser-graphql-wrapper/src/config/index.ts
```diff
@@ -58,4 +58,115 @@ export default {
   },
   parserEndpoint: process.env.PARSER_URL || 'http://example-parser.com/',
   parserRetries: 3,
+  intMask: {
```
kschelonka, line 61: you don't have to do this now, but it's probably worth exporting some kind of config or default setters from the library

kschelonka (approved):

PR #293, opened 2024-03-15
feat(decode): adding in intmask decoding

bassrock:
## Goal

Add in the ability to decode the IntMask so we can turn the encoded Item ID back to a database value for lookup in future queries.


packages/int-mask/src/defaults.ts
```diff
@@ -0,0 +1,105 @@
+import { IntMaskOptions } from './intMask';
+
+export const defaultOptions: IntMaskOptions = {
```
bassrock, line 3: Initial way i had our dummy maps setup were wrong for testing purposes. There are required a certain set of characters and indexes for each number, position.

servers/parser-graphql-wrapper/src/config/index.ts
```diff
@@ -62,78 +62,78 @@ export default {
...
+          ['v', 9],
+          ['w', 9],
+          ['x', 9],
         ]),
     positionMap: process.env.POSITION_MAP
       ? new Map<number, number>(JSON.parse(process.env.POSITION_MAP))
       : new Map<number, number>([
           [9, 0],
-          [123, 2],
+          [12, 1],
```
bassrock, line 122: Numbers must be unique and < 60 on the left side. and right side must have 0-15 defined.

packages/int-mask/src/intMask.ts
```diff
@@ -17,16 +17,52 @@ export type IntMaskOptions = {
 export class IntMask {
   private options: IntMaskOptions;
   private val;
-
   private intMap;
 
-  constructor(val, options: IntMaskOptions) {
+  constructor(val, options?: Partial<IntMaskOptions>) {
```
kschelonka, line 22: the umpteenth time I've wished object defaults in tyepscript were easier

kschelonka (approved):

PR #310, opened 2024-03-15
fix(logs): adding in error logging

bassrock:
## Goal

Add in error log statements to track down a sentry error


servers/list-api/src/businessEvents/sqs/sqsListener.ts
```diff
@@ -40,7 +40,9 @@ export class SqsListener {
       serverLogger.error('unable to add event to queue', {
         data: JSON.stringify(data),
         queueUrl,
+        error: err,
       });
+      Sentry.addBreadcrumb({ message: errorMessage, data: { error: err } });
       Sentry.captureMessage(errorMessage);
```
kschelonka, line 46:
```suggestion
      Sentry.captureException(err);
```

servers/list-api/src/businessEvents/unifiedEventKinesisHandler.ts
```diff
@@ -45,13 +45,15 @@ export class UnifiedEventKinesisHandler {
...
-    } catch {
+    } catch (error) {
       serverLogger.error('Failed to send event(s) to kinesis stream', {
         stream: config.aws.kinesis.unifiedEvents.streamName,
         event: JSON.stringify(unifiedEvent),
+        error,
       });
       Sentry.addBreadcrumb({
         message: `Failed Events: \n ${JSON.stringify(unifiedEvent)}`,
+        data: { error: error },
```
kschelonka, line 56: I think it would be more idiomatic to keep the message as a breadcrumb and change line 60 to be capturing the original error instead.

kschelonka (approved):

PR #314, opened 2024-03-18
feat(esm): terraform modules to esm

bassrock:
## Goal

Packages are starting to only support ESM. This moves terraform modules to ESM so we can get a feel for it.


packages/terraform-modules/src/base/ApplicationAutoscaling.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { Testing } from 'cdktf';
-import { TestResource } from '../testHelpers';
+import { TestResource } from '../testHelpers.js'
```
kschelonka, line 2: I hate this so much

kschelonka (approved):

packages/terraform-modules/src/base/ApplicationAutoscaling.spec.ts
bassrock: so do i

PR #316, opened 2024-03-18
feat(esm): moving account-data-deleter-cdk to esm

bassrock:
## Goal

Move account-data-deleter-cdk to esm


packages/terraform-modules/src/base/ApplicationBaseDNS.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { Testing } from 'cdktf';
-import { TestResource } from '../testHelpers.js'
-import { ApplicationBaseDNS } from './ApplicationBaseDNS.js'
+import { TestResource } from '../testHelpers';
```
bassrock, line 2: Tests dont need to add .js because ts-jest can transpile and module map

PR #318, opened 2024-03-18
feat(v3proxy): v3/send validator and route setup

kschelonka:
I recommend reviewing `servers/v3-proxy-api/src/routes/validations/SendActionValidators.ts` starting from the bottom... Most of the file is not publicly exported.

So... okay.

This code ensures that the actions array passed to the `send` method only contains elements which match one of the schemas defined in the pocket developer api. There are five separate schemas, some of which do not overlap at all (except for the optional time field).

My first cut at writing this was an insanely long decision tree with a lot of nested `if` statements. I thought, 
_These "sub-sanitizers" would really benefit from being treated like scala traits. I bet using a mixin pattern would really clean up this code. Oh good, the typescript docs have an example of how to do mixins although I guess javascript doesn't really handle multiple inheritance._

Is it worse than a huge nested decision tree? No. Is it better? 😬 It's type safe at least!!

You can create an arbitrary sanitizer by building up a chain of validators on a NamedAction base. The outermost class should have a `validate` function which serializes the result to the expected object type. This way we can build up Sanitizers with complete type safety, reusing building blocks along the way...

Example - required fields are oldTag and newTag, optional time field
```
SanitizedTagRename( // serializes internal fields to {oldTag, newTag, time}
  MaybeHasTime( // optionally adds the time field (if valid)
    HasOldTag(  // adds the required oldTag field (if valid)
      HasNewTag(  // adds the required newTag field (if valid)
        NameAction<'tag_rename'> // only applies to {action: 'tag_rename'}
      )
    )
  )
)

// in the editor you'd get type hints on the validation result, 
// and errors if you try to pass any object that doesn't have 
// action: 'add_tag'
```

 
historic-photographs-transform


servers/v3-proxy-api/src/routes/validations/SendActionValidators.ts
```diff
@@ -0,0 +1,662 @@
...
+ * Should be used with a "Sanitize" class to return sanitized
+ * data which conforms to the expected action schema. See `HasItemIdOrUrl`
+ * for example usage.
+ */
+function HasNewTag<TBase extends ActionSanitizable>(Base: TBase) {
+  return class extends Base {
+    readonly newTag: string;
+    constructor(...args: any[]) {
+      super(...args);
+      nonEmptyStringPropOrError(this.input, 'new_tag');
```
bassrock, line 280: 🤣 js. oh js. if only there was a simple not null, not undefined, not empty built in :shakes--fist:

servers/v3-proxy-api/src/routes/validations/SendActionValidators.ts
```diff
@@ -0,0 +1,662 @@
...
+);
+/**
+ * Sanitizer for actions: 'tags_add', 'tags_replace', 'tags_delete'
+ * The input must have a valid item_id or url, the action name, a tags array, and might have
+ * a numeric timestamp.
+ * Calling `ItemTagActionSanitizer(input).validate()` returns a valid
+ * `ItemTagAction` object or throws error if validation fails.
+ */
+export const ItemTagActionSanitizer = SanitizedItemTag(
+  HasItemIdOrUrl(MaybeHasTime(HasValidTags(NamedAction<ItemTagActionNames>))),
```
bassrock, line 577: this is so cool

servers/v3-proxy-api/src/routes/validations/SendActionValidators.ts
```diff
@@ -0,0 +1,662 @@
...
+export const ItemAddActionSanitizer = SanitizedAddItem(
+  HasItemIdOrUrl(MaybeHasTime(MaybeHasTags(MaybeHasTitle(NamedAction<'add'>)))),
+);
+
+/**
+ * Route input to a proper sanitizer based on action name.
+ * This is the main method that should be used from this file.
+ */
+export function ActionSanitizer(input: MaybeAction): SendAction {
+  // Is there a way I can use a map for this without the resulting function
```
bassrock, line 614: 🤔

bassrock (approved):
Wow this is amazing. Really nice work. I like how easy it is to add new send actions if needed.. (hopefully not though)

PR #326, opened 2024-03-21
feat(v3proxy-send): add item property actions to v3/send

kschelonka:
Adds the following actions: archive, delete, favorite, unfavorite

Note that performing these actions with `item_id` as the input will ignore `time` if it is passed, since these mutations do not accept a timestamp (ticket noted).

** Implementation Decisions **

We don't have a batch endpoint for mutating these properties by URL. I decided not to fork the execution for some to be batch and some to be single-item because it would be time-consuming, messy, and totally unnecessary once we have batch endpoints for everything. Instead I'm just using the single-item mutations for both. If we need to optimize we can convert everything to use batch mutations, and the logic will be a lot cleaner.

[POCKET-9632]

[POCKET-9632]: https://mozilla-hub.atlassian.net/browse/POCKET-9632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


servers/v3-proxy-api/src/routes/ActionsRouter.ts
```diff
@@ -119,4 +150,156 @@ export class ActionsRouter {
...
+   *
+   * See `ActionsRouter.archive` for more detailed docstring (same pattern).
+   * @returns true (operation is successful unless error is thrown)
+   * @throws ClientError if operation fails
+   */
+  private async favorite(
+    input: Omit<ItemAction, 'action'> & { action: 'favorite' },
+  ): Promise<true> {
+    if (input.itemId) {
+      // TODO [POCKET-9807]: This mutation does not accept timestamp
```
bassrock, line 212: Hmm i wonder if it should for "true" offline support.

bassrock (approved):

PR #329, opened 2024-03-21
fix(lambda): making lambda ignore git_sha changes

bassrock:
## Goal

Make it so that Lambda deploys happen like ECS deploys in https://github.com/Pocket/pocket-monorepo/pull/324

This should lessen the commenting by the pocket-ci bot unless there is actually a change we should verify.

### DIFF
--- a/.circleci/common.yml
+++ b/.circleci/common.yml
@@ -433,6 +433,7 @@ jobs:
             NEW_ENVVARS=$(aws lambda get-function-configuration --function-name '<< parameters.function-name >>' --query "Environment.Variables | merge(@, \`{\"GIT_SHA\":\"$CIRCLE_SHA1\"}\`)")
             echo $NEW_ENVVARS
             aws lambda update-function-configuration --function-name '<< parameters.function-name >>' --environment "{ \"Variables\": $NEW_ENVVARS }"
+            aws lambda wait function-updated --function-name '<< parameters.function-name >>'
 
             versionId=$(aws lambda publish-version \
                 --function-name '<< parameters.function-name >>' | jq -r .Version)
--- a/.circleci/common.yml
+++ b/.circleci/common.yml
@@ -157,7 +157,6 @@ commands:
       - run: 
           name: Install and setup node
           command: |
-            printenv
             nvm install
             nvm use
             npm install -g pnpm   
@@ -430,6 +429,10 @@ jobs:
                 --s3-key "$s3Key"
 
             aws lambda wait function-updated --function-name '<< parameters.function-name >>'
+
+            NEW_ENVVARS=$(aws lambda get-function-configuration --function-name '<< parameters.function-name >>' --query "Environment.Variables | merge(@, \`{\"GIT_SHA\":\"$CIRCLE_SHA1\"}\`)")
+            echo $NEW_ENVVARS
+            aws lambda update-function-configuration --function-name '<< parameters.function-name >>' --environment "{ \"Variables\": $NEW_ENVVARS }"
 
             versionId=$(aws lambda publish-version \
                 --function-name '<< parameters.function-name >>' | jq -r .Version)
--- a/.circleci/common.yml
+++ b/.circleci/common.yml
@@ -431,7 +431,6 @@ jobs:
             aws lambda wait function-updated --function-name '<< parameters.function-name >>'
 
             NEW_ENVVARS=$(aws lambda get-function-configuration --function-name '<< parameters.function-name >>' --query "Environment.Variables | merge(@, \`{\"GIT_SHA\":\"$CIRCLE_SHA1\"}\`)")
-            echo $NEW_ENVVARS
             aws lambda update-function-configuration --function-name '<< parameters.function-name >>' --environment "{ \"Variables\": $NEW_ENVVARS }"
             aws lambda wait function-updated --function-name '<< parameters.function-name >>'
 
--- a/.circleci/common.yml
+++ b/.circleci/common.yml
@@ -277,9 +277,12 @@ jobs:
           steps:
             - run:
                 name: Terraform apply
+                # Re-add this when tfcmt supports ignoring no change applies
+                # https://github.com/suzuki-shunsuke/tfcmt/issues/1184
+                #  /home/circleci/tfcmt --var target:<< parameters.scope >><<#parameters.dev>>-dev<</parameters.dev>> apply -- terraform apply -auto-approve -lock-timeout=10m
                 command: |
                   cd << parameters.stack-output-path >>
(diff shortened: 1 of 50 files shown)


.circleci/common.yml
```diff
@@ -431,6 +430,10 @@ jobs:
 
             aws lambda wait function-updated --function-name '<< parameters.function-name >>'
 
+            NEW_ENVVARS=$(aws lambda get-function-configuration --function-name '<< parameters.function-name >>' --query "Environment.Variables | merge(@, \`{\"GIT_SHA\":\"$CIRCLE_SHA1\"}\`)")
```
bassrock, line 433: AWS doesn't have a way to update 1 environment variable on a function, so we have to get the existing ones, then replace it with the one we want.

.tfcmt.yml
```diff
@@ -0,0 +1,137 @@
+# Default conifg for us to modify from https://suzuki-shunsuke.github.io/tfcmt/config#default-configuration
+embedded_var_names: []
```
bassrock, line 2: Copying in the template for tfcmt. So we can do a few slight adjustments

.tfcmt.yml
```diff
@@ -0,0 +1,137 @@
...
+      label_color: 1d76db # blue
+      # disable_label: false
+    when_destroy:
+      label: "{{if .Vars.target}}{{.Vars.target}}/{{end}}destroy"
+      label_color: d93f0b # red
+      # disable_label: false
+    when_no_changes:
+      label: "{{if .Vars.target}}{{.Vars.target}}/{{end}}no-changes"
+      label_color: 0e8a16 # green
+      disable_label: true
```
bassrock, line 93: Adjustment 1, no need to comment or label if there are no changes. It clutters up the Github UI

.tfcmt.yml
```diff
@@ -0,0 +1,137 @@
...
+        {{if .Link}}[CI link]({{.Link}}){{end}}
+
+        {{template "guide_apply_parse_error" .}}
+
+        It failed to parse the result.
+
+        <details><summary>Details (Click me)</summary>
+        {{wrapCode .CombinedOutput}}
+        </details>
+# When https://github.com/suzuki-shunsuke/tfcmt/issues/1184 is done, we can make it so tfcmt doesn't comment on a no-op apply!
```
bassrock, line 137: Future change 2, We should only get the "apply" comment when terraform actually does something. But its pending work in the tfcmt repo. So for now, we are just gonna turn off apply by not putting it through tfcmt. 

If an apply fails, the ci job will fail and we can look at the CI logs for now.

infrastructure/account-data-deleter/src/lambda/base/sqsLambda.ts
```diff
@@ -29,7 +29,7 @@ export class SqsLambda extends Construct {
   ) {
     super(scope, name.toLowerCase());
 
-    const { sentryDsn, gitSha } = this.getEnvVariableValues();
+    const { sentryDsn } = this.getEnvVariableValues();
```
bassrock, line 32: We need to stop passing GitSha through via Terraform. Like ECS this is a variable that gets updated via "code deployment" and we ignore in our infrastructure.

packages/terraform-modules/src/base/ApplicationLambdaCodeDeploy.ts
```diff
@@ -43,7 +44,19 @@ export class ApplicationLambdaCodeDeploy extends Construct {
   ) {
     super(scope, name);
 
-    this.codeDeployApp = this.setupCodeDeploy();
+    const { codeDeployApp, codeDeployGroup } = this.setupCodeDeploy();
+    this.codeDeployApp = codeDeployApp;
+    this.codeDeployGroup = codeDeployGroup;
+
+    new TerraformOutput(this, 'lambda_codedeploy_app', {
```
bassrock, line 51: Start outputing the variables for us to use if we wanted.

However, using them will be a seperate future followup. Because we use multiple lambdas in each CI we can't use a "static" id like we do in ECS. So need to think through the best way to dynamically pull the right ones.

packages/ts-logger/src/logger.ts
```diff
@@ -75,7 +75,7 @@ export function createLogger(options: LoggerOptions | undefined = {}): Logger {
   };
 
   const enchancedDefaultMeta = {
-    releaseSha: process.env.RELEASE_SHA || null,
+    releaseSha: process.env.RELEASE_SHA || process.env.GIT_SHA || null,
```
bassrock, line 78: Sometimes the sha is RELEASE_SHA (ECS, DOCKER), sometimes its GIT_SHA, for now lets try both and we can clean this up in future prs.

.circleci/common.yml
```diff
@@ -278,9 +277,12 @@ jobs:
           steps:
             - run:
                 name: Terraform apply
+                # Re-add this when tfcmt supports ignoring no change applies
+                # https://github.com/suzuki-shunsuke/tfcmt/issues/1184
+                #  /home/circleci/tfcmt --var target:<< parameters.scope >><<#parameters.dev>>-dev<</parameters.dev>> apply -- terraform apply -auto-approve -lock-timeout=10m
                 command: |
                   cd << parameters.stack-output-path >>
-                  /home/circleci/tfcmt --var target:<< parameters.scope >><<#parameters.dev>>-dev<</parameters.dev>> apply -- terraform apply -auto-approve -lock-timeout=10m
+                  terraform apply -auto-approve -lock-timeout=10m
```
kschelonka, line 285: Does this throw error the way it was done previously?

bassrock: Yea the CI job will fail. Just won't comment on the PR for now.

packages/terraform-modules/src/base/ApplicationLambdaCodeDeploy.ts
kschelonka: can you expand on what you mean by "multiple lambdas in each CI"?

packages/terraform-modules/src/pocket/PocketApiGatewayLambdaIntegration.ts
```diff
@@ -84,9 +84,8 @@ export class PocketApiGateway extends Construct {
       route.lambda.lambda.versionedLambda,
     ]);
 
-    // Use the current timestamp in milliseconds to trigger a redeployment if no triggers are provided.
-    // If no triggers are provided, this resource will be redeployed on every "terraform apply".
-    const triggers = config.triggers ?? { deployedAt: Date.now().toString() };
+    // If no triggers are provided, this resource will be redeployed only if something changes to cause it.
+    const triggers = config.triggers ?? {};
```
kschelonka, line 88: 👍

kschelonka (approved):

PR #346, opened 2024-03-25
fix(tfenv): make tfenv cache with a changing cache key

bassrock:
## Goal

Every job was always downloading Terraform instead of reusing the cache. This will cache terraform based on the version defined in the `.terraform-version`.


.circleci/common.yml
```diff
@@ -285,16 +285,31 @@ jobs:
       - exit-early-if-irrelevant:
           for: << parameters.for >>
       - checkout
-      - restore_cache:
-          name: Restore Tfenv
-          keys:
-            - tfenv-v2
+      - unless:
+          # When it is not raw hcl, the stack output path is 3 folders away from our .terraform-version
+          condition: <<parameters.uses_raw_hcl>>
```
kschelonka, line 290: is this parameter already set?

kschelonka (approved):

.circleci/common.yml
bassrock: yea it is

PR #356, opened 2024-03-25
fix(apns2): upgrading to APNS2

bassrock:
## Goal

node-apns is deprecated and not maintained anymore. Furthermore it has security vulnerabilities in the underlying node-forge library.

This upgrades us to the latest apns2 library that is used in the community, most things are all a 1:1 since the APNS api is not any different just library.

### DIFF
--- a/servers/push-server/src/apns.ts
+++ b/servers/push-server/src/apns.ts
@@ -1,52 +1,59 @@
 import {
-  NotificationAlertOptions,
-  ProviderToken,
-  ResponseFailure,
-  Responses,
-  Provider,
+  ApnsClient,
+  Errors,
   Notification,
-  token as apnToken,
-} from 'apn';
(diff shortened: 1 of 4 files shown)


servers/push-server/src/apns.ts
```diff
@@ -1,52 +1,59 @@
...
-    note.payload = payload;
+    const noteOptions: NotificationOptions = {
+      expiration: Math.floor(Date.now() / 1000) + 3600, // Expires 1 hour from now.
+    };
 
     if (isSilent) {
-      note.pushType = 'background';
-      note.priority = 5;
-      note.contentAvailable = true;
+      noteOptions.type = PushType.background;
```
kschelonka, line 48: human readable!

kschelonka (approved):
consider adding smoke test if not a big lift

PR #361, opened 2024-03-26
feat(list): add tag mutations accessed by name

kschelonka:
To support v3 proxy /send endpoint, which can identify a SavedItem either by URL or ID interchangeably, and uses tag names instead of tag IDs.

Adds:
* clearTags - clear all tags from a saved item referenced by id or url
* renameTagByName - rename a tag by its name (not id)
* deleteTagByName - delete a tag by its name (not id)
* removeTagsByName - remove tags from a saved item by names, not ids; saveditem referenced by id or url
* replaceTags - replace all tags on a saved item referenced by id or url

### DIFF
--- a/servers/list-api/schema.graphql
+++ b/servers/list-api/schema.graphql
@@ -912,7 +912,7 @@ type Mutation {
 
 """
 We don't have official oneOf support, but this will
-throw if one of these fields is not set and not null.
+throw if both `id` and `url` are unset/null.
 Don't provide both... but if both are provided, it will
 default to using ID.
 """
--- a/servers/list-api/schema.graphql
+++ b/servers/list-api/schema.graphql
@@ -871,6 +871,51 @@ type Mutation {
     timestamp: ISOString!
     title: String! @constraint(minLength: 1, maxLength: 75)
   ): SavedItem
+
+  """
+  Remove all tags associated to a SavedItem (included for v3 proxy).
+  """
+  clearTags(savedItem: SavedItemRef!, timestamp: ISOString): SavedItem
+    @tag(name: "v3proxy")
+  """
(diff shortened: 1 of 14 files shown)


servers/list-api/schema.graphql
```diff
@@ -871,6 +871,54 @@ type Mutation {
...
+  """
+  replaceTags(
+    savedItem: SavedItemRef!
+    tagNames: [String!]!
+    timestamp: ISOString
+  ): SavedItem @tag(name: "v3proxy")
+}
+
+"""
+We don't have official oneOf support, but this will
```
bassrock, line 914: one day :(

servers/list-api/src/resolvers/mutation.ts
```diff
@@ -319,3 +321,90 @@ export async function updateTag(
...
+  context: IContext,
+): Promise<SavedItem> {
+  if (!atLeastOneOf(args.savedItem, ['id', 'url'])) {
+    throw new UserInputError('SavedItemRef must have one of `id` or `url`');
+  }
+  // Previously clients have used this to clear tags by passing an
+  // empty replacement array, so reroute to clearTags mutation if
+  // if the array of tagNames is empty
+  if (args.tagNames.length === 0) {
+    return clearTags(root, args, context);
```
bassrock, line 338: 🔥

servers/list-api/src/types/index.ts
```diff
@@ -282,3 +282,5 @@ export type SaveUpdateTagsInputDb = {
 };
 
 export type SaveByIdResult = NotFound | PocketSave;
+
+export type SavedItemRefInput = { id?: string; url?: string };
```
bassrock, line 286: one. day. we. will. get. oneOf.

bassrock (approved):

PR #369, opened 2024-03-27
feat(v3proxy:tagging): implement single-item tag actions

kschelonka:
Adds v3/send actions for tag operations on a single item (adding tags, removing tags, replacing tags, clearing tags)

Related change:
* Add default timestamp to sanitizers now that all mutations are updated to accept timestamp (some required, some optionally to be backwards-compatible). There is no reason to defer to the subgraph for the timestamp, so let's just remove some conditionality and populate it when the actions are validated.

[POCKET-9622]

[POCKET-9622]: https://mozilla-hub.atlassian.net/browse/POCKET-9622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bassrock (approved):
LGTM

PR #374, opened 2024-03-27
feat(instant-sync): adding in all the necessary instant sync scaffolding

bassrock:
## Goal

Add in instant_sync event scaffolding. 

Instant sync is a feature where a silent push notification is sent to Android and iOS when a user does something on their saves, and triggers a sync. 

Today this occurs on an action to the v3 api but not on GraphQL. 

This PR aims to listen to list events occurring in event bridge and just log the count for now.

All logic will be in followups

https://mozilla-hub.atlassian.net/browse/POCKET-9810


infrastructure/instant-sync-events/src/sqsEventLambda.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  constructor(scope: Construct, name: string, config: SqsLambdaProps) {
+    super(scope, name.toLowerCase());
+
+    const { sentryDsn } = this.getEnvVariableValues();
+
+    this.construct = new PocketSQSWithLambdaTarget(this, name.toLowerCase(), {
+      name: `${stackConfig.prefix}-${name}`,
+      batchSize: 100,
+      batchWindow: 30,
```
kschelonka, line 31: 👍

kschelonka (approved):

PR #376, opened 2024-03-28
feat(lambda): adding in the ability to add the lambda secret layer

bassrock:
## Goal

* Add [Lambda Secret Layer support to terraform modules](https://docs.aws.amazon.com/secretsmanager/latest/userguide/retrieving-secrets_lambda.html)
* Build a helper [package for fetching a secret from the lambda secret extension](https://docs.aws.amazon.com/systems-manager/latest/userguide/ps-integration-lambda-extensions.html)


kschelonka (approved):
🍰

PR #381, opened 2024-03-29
fix(v3proxy:id): media response for /get and /fetch uses resolvedId

kschelonka:
To be consistent with /v3 api, use resolvedId
instead of itemId on media response (videos, images, etc.). The field is called 'item_id' which is confusing , and commonly the resolved_id is the same as item_id, but it is indeed resolved_id.

[POCKET-9677]

[POCKET-9677]: https://mozilla-hub.atlassian.net/browse/POCKET-9677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bassrock (approved):

servers/v3-proxy-api/src/test/fixtures/fetch.ts
```diff
@@ -14,8 +14,8 @@ export const expectedFetch: FetchResponse = {
   },
   cachetype: 'db',
   list: {
-    '11231399273': {
-      item_id: '11231399273',
+    '2111834840': {
```
kschelonka, line 17: just changed the item_id of one of the fixtures (and left the resolved_id alone)

PR #382, opened 2024-03-29
fix(sync): adding instant sync logic

bassrock:
# Goal

Enable instant sync based on our event stream from the event bridge.

# Implementation Notes

There is some extra logic in the old web repo around a user_tokens table, we have really stopped using that table years ago, so I did not port its logic here.

Once this is deployed, we can also remove all the push logic from web repo around this.


servers/push-server/src/gcm.ts
```diff
@@ -8,22 +8,12 @@ const app = initializeApp({
...
-}
-
 export const sendNotificationToDevice = (
   notification: any,
-  payload: Payload,
   token: any,
 ): Promise<string> => {
   const message: Message = {
     token: token,
-    data: {
```
bassrock, line 23: Clients no longer use these.

servers/push-server/src/worker.ts
```diff
@@ -23,40 +19,13 @@ const processMessage = async (fullMessage: Message): Promise<void> => {
     level: 'info',
   });
 
-  if (target === TARGET_APNS) {
```
bassrock, line 26: No longer used

servers/push-server/src/worker.ts
```diff
@@ -23,40 +19,13 @@ const processMessage = async (fullMessage: Message): Promise<void> => {
...
-      token,
-      false,
-    );
-  } else if (target === TARGET_APNS_SILENT) {
+  if (target === TARGET_APNS_SILENT) {
     console.log('APNS SILENT push', token);
-    await apns.sendNotificationToDevice(
-      contents,
-      {
-        g: message.guid,
```
bassrock, line 44: Clients don't rely on the guid payload anymore. (verified in android and ios codebase)

bassrock: It also simplifies the required code here.

lambdas/instant-sync-events/src/clients.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Create a db client for reads from readitla_ril-* databases
+ */
+export async function readClient(): Promise<knex.Knex<any, any[]>> {
+  if (readDb) return readDb;
+
+  const secret = await fetchSecret(config.dbSecretName);
+
+  readDb = createConnection({
+    host: secret.read_host,
+    port: 3306,
```
kschelonka, line 18: is this not in the secret?

lambdas/instant-sync-events/src/handlerFn.ts
```diff
@@ -0,0 +1,142 @@
...
+  'FAVORITE_ITEM',
+  'UNFAVORITE_ITEM',
+  'ARCHIVE_ITEM',
+  'UNARCHIVE_ITEM',
+  'ADD_TAGS',
+  'REPLACE_TAGS',
+  'CLEAR_TAGS',
+  'REMOVE_TAGS',
+  'RENAME_TAG',
+  'DELETE_TAG',
```
kschelonka, line 23: 'UPDATE_TITLE' - new event

lambdas/instant-sync-events/src/clients.ts
bassrock: no lol idk i feel a db port is alweays the same

kschelonka (approved):

PR #383, opened 2024-03-30
feat(v3proxy:annotations): optionally add annotations to v3/get response

kschelonka:
~Teeeechnically I slacked off on some test coverage checking for the options to be propagated through the function stack from the request body...~

un-slacked the coverage and found a bug where I missed passing the options to a line that got broken up by the linter, tl;dr kat should never not use tdd

but the transformers are unit-tested

** related changes **
* shorten some of the query names (sorry about the diffs)


bassrock (approved):
LGTM, we have very different versions of slacked off on test coverage 🤣

PR #405, opened 2024-04-04
fix(errors): cleanup the error logs

bassrock:
## Goal

Our logs are filled with unhelpful stack traces that provide too much data.

This removes the stack from the server logs.

## Parser
Separately we need some more data on the parser to better debug issues we are currently seeing.

### DIFF
--- a/servers/parser-graphql-wrapper/src/dataLoaders/articleLoader.ts
+++ b/servers/parser-graphql-wrapper/src/dataLoaders/articleLoader.ts
@@ -41,7 +41,12 @@ export const getArticleByUrl = async (
   const data = await new FetchHandler().fetchJSON(endpoint);
   // check if there's an item
   if (!data || (data && !data.item)) {
-    Sentry.captureException(new Error(`No item found for URL: ${url}`));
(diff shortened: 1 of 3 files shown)


servers/parser-graphql-wrapper/src/dataLoaders/articleLoader.ts
```diff
@@ -41,7 +41,12 @@ export const getArticleByUrl = async (
   const data = await new FetchHandler().fetchJSON(endpoint);
   // check if there's an item
   if (!data || (data && !data.item)) {
-    Sentry.captureException(new Error(`No item found for URL: ${url}`));
+    Sentry.addBreadcrumb({
+      level: 'debug',
+      message: 'article loader parser request',
+      data: { originalUrl: url, endpoint },
+    });
+    Sentry.captureException(new Error(`No item found for URL`));
```
bassrock, line 49: using the same error message ensures sentry will group these together, and we can use the breadcrumbs to understand.

kschelonka (approved):

PR #412, opened 2024-04-05
feat(v3): adding v3 premium fields

bassrock:
## Goal

Add in needed premium fields that V3 needs to provide

https://mozilla-hub.atlassian.net/browse/POCKET-9916

### DIFF
--- a/servers/user-api/src/resolvers/index.ts
+++ b/servers/user-api/src/resolvers/index.ts
@@ -10,7 +10,7 @@ import {
   updateUserEmailByFxaId,
   expireUserWebSessionByFxaId,
 } from './mutations';
-import { User } from '../types';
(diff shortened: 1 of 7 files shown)


servers/user-api/src/resolvers/index.ts
```diff
@@ -91,6 +91,32 @@ export const resolvers = {
...
+    premiumFeatures: async (
+      parent,
+      args,
+      context: IContext,
+    ): Promise<PremiumFeature[] | null> => {
+      const status =
+        (await parent.premiumStatus) ??
+        (await context.models.user.premiumStatus);
+      if (status !== PremiumStatus.ACTIVE) {
+        return null;
```
bassrock, line 110: Im torn on null or empty array, but went null cause it should be an easier client check.

kschelonka (approved):

PR #420, opened 2024-04-08
fix(fetch): default annotations to on for fetch

bassrock:
## Goal

Android does not fetch annotations by default on initial login, but should so users get their highlights.

https://mozilla-hub.atlassian.net/browse/POCKET-9924


nzeltzer (approved):
lgtm

PR #432, opened 2024-04-08
fix(sort): adding in longest and shortest sort no-op

bassrock:
## goal

Allow android requests on get for shortest and longest sort order to succeed


kschelonka (approved):
🆕

PR #439, opened 2024-04-09
fix(cache): moving to a much lower serverless redis cache

bassrock:
## Goal

Lower costs of the parser graphql wrapper redis by limiting the serverless cache to 7GB. This must be done as a new cache, because of how terraform works with serverless redis.


nzeltzer (approved):
lgtm

PR #465, opened 2024-04-16
chore(circleci): optimizing costs

bassrock:
## Goal

It looks like we are using 1.1 TB of egress network costs due to caching within our hosted runner. This removes all caching within the CircleCI hosted runner for now to save on egress network costs which make up more then half the circleci bill.

This also disables docker layer caching which is also way above our maximum for now as well.


nzeltzer (approved):
Lgtm

PR #468, opened 2024-04-16
chore(pnpm): updating to pnpm 9

bassrock:
## Goal

PNPM 9 came out and we were not locked in a few places, might as well update!


collectedmind:
giphy

PR #472, opened 2024-04-17
fix(tests): making the tests not change the time

bassrock:
## Goal

When I moved the list api and other services, I moved them and replaced sinon with pure jest. There was some weird behavior with the new timers behavior in Jest that I could not fully sort out at the time that caused us to have to set `advanceTimers` to `true`.  This had the side effect of causing our tests that check dates to be a bit flakey because the time could increment in the amount of time it took to run the test instead of being 

Upon looking into the behavior `advanceTimers` was needed because the new timer behavior fakes async tick functions that we don't want it to. See https://github.com/nock/nock/issues/2200#issuecomment-1280957462

By telling jest to not fake the async node functions, the clock stays frozen and the rest of the code can execute.


kschelonka (approved):
⏳

PR #483, opened 2024-04-18
fix(user): fix query for retrieving premium active status

kschelonka:
Fixes the SQL query for returning premium subscription (see comments for explanation).

[POCKET-9980]

[POCKET-9980]: https://mozilla-hub.atlassian.net/browse/POCKET-9980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/servers/user-api/src/dataService/userDataService.ts
+++ b/servers/user-api/src/dataService/userDataService.ts
@@ -142,17 +142,20 @@ export class UserDataService {
         'users.user_id',
       )
       .leftJoin(
-        this.readDb.raw(`(
-        SELECT user_id,
-               CASE
-                   WHEN active = 1 THEN 'ACTIVE'
-                   WHEN active = 0 THEN 'EXPIRED'
-                   ELSE 'NEVER'
-               END AS premiumStatus
-        FROM payment_subscriptions
-        ORDER BY active DESC, expires_at DESC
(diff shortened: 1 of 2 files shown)


servers/user-api/src/dataService/userDataService.ts
```diff
@@ -142,17 +142,20 @@ export class UserDataService {
         'users.user_id',
       )
       .leftJoin(
-        this.readDb.raw(`(
-        SELECT user_id,
```
kschelonka, line 146: this was just grabbing the first row returned by the ordering (expires_at desc, active desc) regardless of user_id, and limiting to 1

so the subsequent join would return null (and set the user to be 'NEVER')

servers/user-api/src/test/functional/user.integration.ts
```diff
@@ -266,6 +266,13 @@ describe('User', () => {
 
     it('should load active subscription status', async () => {
       await writeDb('payment_subscriptions').insert([
+        {
```
kschelonka, line 269: add a more recent value to repro the bug

bassrock (approved):

PR #486, opened 2024-04-19
feat(v3proxy): add account details to proxy response

kschelonka:
If account/forceaccount=1, return account details on /v3/get

_Edit: these are wontdos per Daniel's PR comments_
~Tabled to follow-up tickets due to not being available in the graph:~
* ~retrieving all aliases (currently just transform the email we have as the only alias and set it to 'confirmed')~
* ~retrieving if the user's premium status is going to expire (for upsell prompt)~

[POCKET-9917]

[POCKET-9917]: https://mozilla-hub.atlassian.net/browse/POCKET-9917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


servers/v3-proxy-api/src/graph/get/toRest.ts
```diff
@@ -99,6 +105,92 @@ function searchStatusResponse(
...
+      email: accountData.email,
+      birth: mysqlTimeString(
+        new Date(accountData.accountCreationDate),
+        'US/Central',
+      ),
+      first_name: firstName,
+      last_name: lastName,
+      premium_status: accountData.isPremium ? '1' : '0',
+      is_fxa: accountData.isFxa ? 'true' : 'false',
+      // TODO - Fetch all aliases POCKET-9882
```
bassrock, line 128: @kschelonka moving forward with our FxA work we disabled creation of new aliases. I think its ok to not do this.

servers/v3-proxy-api/src/graph/get/toRest.ts
```diff
@@ -99,6 +105,92 @@ function searchStatusResponse(
...
+        type: 'pocket' as const, // static value
+        sort_id: 1, // static value
+      },
+      premium_features: accountData.premiumFeatures.map((feat) =>
+        PremiumFeatureTransformer(feat),
+      ),
+      premium_alltime_status: AlltimeStatusTransformer(
+        accountData.premiumStatus,
+      ),
+      premium_on_trial: '0', // TODO POCKET-9981 - don't have this value
```
bassrock, line 153: @kschelonka fine to skip. We are not actively using it.

bassrock (approved):

PR #494, opened 2024-04-23
chore(kysley): switch parser-wrapper to keysley

bassrock:
# Goal

This switches the parser wrapper to use keysly.


servers/parser-graphql-wrapper/package.json
```diff
@@ -12,7 +12,9 @@
     "test-integrations": "jest \"\\.integration\\.ts\" --runInBand --force-exit",
     "lint": "eslint --fix-dry-run \"src/**/*.ts\"",
     "format": "eslint --fix \"src/**/*.ts\"",
-    "prebuild": "graphql-codegen"
+    "prebuild": "graphql-codegen",
+    "generate:database:types:readitlab": "DATABASE_URL=mysql://root@localhost:3306/readitla_b kysely-codegen --include-pattern=\"readitla_b.items_resolver\" --out-file=src/__generated__/readitlab.d.ts",
```
kschelonka, line 16: just a note here that since this isn't the production database, there is risk that it will become out-of-sync

kschelonka (approved):
ahh kysely is great

servers/parser-graphql-wrapper/package.json
bassrock: talked via slack and agreed this is ok for now.

PR #497, opened 2024-04-23
fix(ids): updating id logic to use a hash

bassrock:
## Goal

Update the id on the items to include a md5 hash that we validate so users can't do any random strings.


servers/parser-graphql-wrapper/src/datasources/ParserAPI.ts
```diff
@@ -146,7 +146,7 @@ export class ParserAPI extends RESTDataSource {
   private parserResponseToItem(parserResponse: ParserResponse): Item {
     return {
       itemId: parserResponse.item_id,
-      id: IntMask.encode(parserResponse.item_id),
+      id: createItemId(parserResponse.item_id),
```
kschelonka, line 149: @collectedmind I know we have a million ids... should we add `readerViewId` to make it more obvious what this is...

PR #522, opened 2024-04-30
feat(shares): edit share context after creation

kschelonka:
To speed up the experience of creating links;
create the link on one view with a mutation, then optionally
use this mutation to attach context.

This is intended as a step done immediately after sharing the link, not coming back to it way later. This is why we can use the guid. It also lets us enable logged-out users to share links and have the same behavior, if we want to in the future.

** Changes **
* Created salts for userId (unused, but maybe in the future) and guid; uploaded to aws prod and dev accounts
* Save hashed + salted userId + guid (if available from JWT) when creating shares -- this future-proofs if we want to look up by the hashed userid in the future; note this is a one-way lookup, not an encryption/masking
* Add mutation to attach context to a link - checks hashed guid for ownership (same hashed guid used to create the record)
* Returns generic not found if the ownership check failed, or if the link does not exist

[POCKET-10035]

[POCKET-10035]: https://mozilla-hub.atlassian.net/browse/POCKET-10035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/servers/shares-api/src/apollo/context.ts
+++ b/servers/shares-api/src/apollo/context.ts
@@ -6,7 +6,8 @@ import {
   SharesDataSourceUnauthenticated,
 } from '../datasources/shares';
 import { dynamoClient } from '../datasources/dynamoClient';
-
+import { UserContext, UserContextFactory } from '../models/UserContext';
+import { config } from '../config';
 /**
  * Context factory function. Creates a new context upon
  * every request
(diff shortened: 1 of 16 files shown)


servers/shares-api/src/apollo/context.ts
```diff
@@ -25,32 +26,46 @@ export async function getContext({
...
   PocketShareModel: PocketShareModel;
+  User: UserContext;
   constructor(options: { request: Request }) {
-    const userId = options.request.headers.userid;
+    const rawUserId = options.request.headers.userid;
+    const rawGuid = options.request.headers.guid;
     const isNative =
       options.request.headers.applicationisnative === 'true' ? true : false;
     const client = dynamoClient();
+    // We have an authenticated user?
```
bassrock, line 41: Weeee! so many conditionals.... prob should be a package... at some point :p

bassrock (approved):

servers/shares-api/src/apollo/context.ts
kschelonka: Yeah would be good to just have a JWT user model that's shared

PR #523, opened 2024-05-01
feat(v3proxy:errors): skip Forbidden errors orginating from recentSearch

kschelonka:
Handle partial error/data response, since we still want the rest of the data even if recentSearch returns null + Forbidden error (i.e. if the requester is not logged into a premium account). Supports Android pattern for retrieving recent search history.

[POCKET-10034]

This was tested manually by making real calls to the graph using a premium and non-premium account. I'm trying to figure out if it's possible to test this another way, but I can't mock the `GraphQLClient.request` method anymore because it will not invoke the middleware... Alternative is standing up a local graph server that mirrors prod or making canary requests to the real graph...

[POCKET-10034]: https://mozilla-hub.atlassian.net/browse/POCKET-10034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/servers/v3-proxy-api/src/graph/graphQLClient.ts
+++ b/servers/v3-proxy-api/src/graph/graphQLClient.ts
@@ -1,5 +1,17 @@
-import { GraphQLClient } from 'graphql-request';
-
+import {
+  ClientError,
+  Variables,
+  RequestDocument,
+  GraphQLClient,
+  resolveRequestDocument,
+} from 'graphql-request';
+import {
+  GraphQLRequestContext,
+  GraphQLResponse,
+  ResponseMiddleware,
+  VariablesAndRequestHeadersArgs,
+} from 'graphql-request/build/esm/types';
+import type { TypedDocumentNode } from '@graphql-typed-document-node/core';
 import {
   SavedItemsSimpleQuery,
   SavedItemsSimpleQueryVariables,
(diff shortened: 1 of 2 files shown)


servers/v3-proxy-api/src/graph/graphQLClient.ts
```diff
@@ -18,55 +30,154 @@ import {
...
+  constructor(
+    accessToken: string,
+    consumerKey: string,
+    private headers: any,
+  ) {
+    //these headers are not compatible with GraphQLClient's fetch.
+    //they throw an error instead, so ignoring them
+    // headers might include cookie, some observability traceIds, apollo studio headers
+    // so, we are implicitly passing them and removing only those that causes issues
+    delete headers['content-length'];
```
kschelonka, line 83: this is basically the old `getClient` function but the diff is weird

servers/v3-proxy-api/src/graph/graphQLClient.ts
```diff
@@ -18,55 +30,154 @@ import {
...
+    context: GraphQLRequestContext,
+  ): ResponseMiddleware {
+    const middleware: ResponseMiddleware = (response) => {
+      if (!(response instanceof Error) && response.errors) {
+        // Filter out the recent searches forbidden error
+        // In the proxy this data is requested regardless of
+        // premium status, so it's just skipped
+        // If we want to add more matchers for skipping other
+        // errors we can refactor this to become more modular
+        const errors = response.errors.filter(
```
kschelonka, line 153: could unit-test this

bassrock (approved):
LGTM. I really appreciate all the comments here

PR #529, opened 2024-05-02
feat(es6): convert to es6

bassrock:


bassrock:
closing this. Discussed with @kschelonka and we will do the following instead:

* maker our eslint stricter to:
   * require .js
   * require importing jest in files
* remove node-fetch for native fetch
* in our own code:
   * remove default exports for object exports

PR #531, opened 2024-05-03
fix(v3proxy:tags): return tags on detailType=simple

kschelonka:
Previously tags were only returned on
detailType=complete

[POCKET-10025]

fix(v3proxy:error): fix accessing undefined path at runtime

When checking errors, it seems like path is not always defined on the Graphql client despite type definition.

Should fix https://pocket.sentry.io/issues/5298029357/?project=4504658424102912&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=24h&stream_index=0

[POCKET-10025]: https://mozilla-hub.atlassian.net/browse/POCKET-10025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


servers/v3-proxy-api/src/test/fixtures/fetch.ts
```diff
@@ -107,11 +107,11 @@ export const expectedFetch: FetchResponse = {
       },
       tags: {
         research: {
-          item_id: '11231399273',
+          item_id: '2111834840',
```
kschelonka, line 110: for some reason this was set to be the resolved_id, but it should definitely be the item_id...?

bassrock (approved):

PR #532, opened 2024-05-06
feat: enable custom ORM tracing instrumentations per service

jpetto:
- opens up this package to services using prisma

i am lost without a PR template 🙃 

anything more y'all would like to see here? tests for constructing the instrumentation array? a push to dev?


bassrock (approved):
LGTM!

Can you make sure knex is added to all the servers that use it? I think a few are missing

bassrock (approved):

PR #540, opened 2024-05-08
feat(eslint): upgrading to eslint9

bassrock:
## Goal

Implements ESLint 9 using the new FlatPack.

in doing this I did remove the eslint-import plugins since they do not support FlatPack yet. Also upon reflection I think those types of errors should be caught by tsconfig build errors, not eslint.

I also removed our extra prettier rules, upon reading their documentation it seems that we should be covered by just using the plugin and it seems to hold up in my testing.


packages/eslint-config/base.js
```diff
@@ -0,0 +1,41 @@
+// @ts-check
+
+import eslint from '@eslint/js';
+import tseslint from 'typescript-eslint';
+import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended';
+
+export default tseslint.config(
+  eslint.configs.recommended,
+  ...tseslint.configs.recommended,
+  // TODO: Enable after the change to eslint9 and fix errors in a followup
```
kschelonka, line 10: ticketed?

kschelonka (approved):

packages/eslint-config/base.js
bassrock: yea ill make a bunch of tickets, i think we need to do it 1 by 1

PR #546, opened 2024-05-09
fix(hoisting): remove all package hoisting

bassrock:
## Context

Right now we do not get accurate import or package missing errors, because we are hoisting all packages to the top level. Which... is a bad no no.  The reason we did this is because it was difficult to keep all the packages defined in multiple places at the same version.  

## Goal
Remove all package hoisting, utilizing Syncpack to ensure that all packages in the repo are the same version. This will ensure that we get compile time errors when we are missing packages.

## Todo
- [x] remove hoisting
- [x] go package by package adding missing ones.
- [x] re-order all package.json files with sort-package-json so that they all look similar.


kschelonka (approved):
💪

PR #553, opened 2024-05-10
chore(search): add and use auto-generated types

kschelonka:
Adds auto-generated typing to user-list-search in preparation for doing more work in this repository.

Remove the manually generated types that are no longer necessary.

Remove a spec test that covers an impossible case.


servers/user-list-search/src/datasource/elasticsearch/searchQueryBuilder.spec.ts
```diff
@@ -111,13 +111,13 @@ describe('SearchQueryBuilder', () => {
       },
       {
         name: 'contentType',
-        filter: { contentType: SavedItemContentType.IMAGE },
+        filter: { contentType: SearchItemsContentType.Video },
```
kschelonka, line 114: image type exists in elasticsearch index but is not exposed in the api

servers/user-list-search/src/resolvers.spec.ts
```diff
@@ -1,27 +0,0 @@
...
-    };
-
-    const badSearch = async () => {
-      return await resolvers.User.search(
-        {},
-        {
-          params: inputParams,
-        },
-        {
-          userId: undefined,
```
kschelonka, line 17: impossible case

bassrock (approved):

PR #567, opened 2024-05-15
feat(search:corpus): add indices for language-specific corpora

kschelonka:
Set up local infrastructure and environment variables for corpus search indices. These indices are language-specific. 

**Note that this doesn't actually create the indices on the prod/dev clusters**, but it does locally. I'm not sure how we want to capture this. In the `bin` folder? It's kind of like a schema migration.

Related changes:
* reduce the shards in the local list index

Related PR:
* enrich event bridge events with additional data: https://github.com/Pocket/content-monorepo/pull/165

[POCKET-10121]

[POCKET-10121]: https://mozilla-hub.atlassian.net/browse/POCKET-10121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.docker/aws-resources/elasticsearch/esindex_corpus_de.json
+++ b/.docker/aws-resources/elasticsearch/esindex_corpus_de.json
@@ -1,0 +1,69 @@
+{
+  "settings": {
+    "index": {
+      "number_of_shards": 1,
+      "number_of_replicas": 1
+    }
+  },
+  "mappings": {
+    "_source": {
+      "enabled": true
+    },
+    "properties": {
+      "id": {
+        "type": "keyword"
+      },
+      "title": {
+        "type": "text",
+        "analyzer": "german"
+      },
+      "url": {
+        "type": "text"
+      },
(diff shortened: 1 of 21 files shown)


.docker/aws-resources/elasticsearch/esindex_corpus_de.json
```diff
@@ -0,0 +1,69 @@
...
+    "_source": {
+      "enabled": true
+    },
+    "properties": {
+      "id": {
+        "type": "keyword"
+      },
+      "title": {
+        "type": "text",
+        "analyzer": "german"
```
kschelonka, line 18: these language-specific analyzers have built-in stemming, tokenization, etc.

bassrock (approved):
Lets treat the indexes as scripts we have to run on VPN for now. if we do this regualrly we can automate

PR #575, opened 2024-05-22
feat(exploreCorpus): process collection and corpus events

kschelonka:
Set up terraform infrastructure for indexing new collections and corpus items in elasticsearch cluster. Sets up rules for forwarding relevant events to an SNS topic, which fans out to SQS queue that is batch-consumed by a Lambda. This lambda makes bulk index requests to elasticsearch cluster.

Contains the app code for the lambda indexer.

Updates the language-specific corpora indices -- remove a duplicative ID field and add a parent_id field for collection stories, which points to the parent collection.

Draft as I test in dev -- need to do some manual testing


infrastructure/user-list-search/sqs_sns_topic_subscription.tf
```diff
@@ -19,6 +28,30 @@ resource "aws_sns_topic_subscription" "user_events_sns_topic_subscription" {
...
+  #    deadLetterTargetArn: aws_sqs_queue.corpus_events_sns_topic_dlq.arn
+  #  })
+  depends_on = [aws_sqs_queue.corpus_events_sns_topic_dlq, aws_sqs_queue.corpus_events]
+}
+
+resource "aws_sns_topic_subscription" "collection_events_sns_topic_subscription" {
+  topic_arn = local.collectionEventsSnsTopicArn
+  protocol  = "sqs"
+  endpoint  = aws_sqs_queue.corpus_events.arn
+  # The version of terraform used in this service does not support redrive_policy
```
bassrock, line 47: Hmm.. this should be on 1.8..

bassrock (approved):
You got deep with the infr. Lets backlog at ticket on cleaning up the search lambdas so future us doesnt have to deal with that

infrastructure/user-list-search/sqs_sns_topic_subscription.tf
kschelonka: I copy-pasted this, maybe can be removed?

PR #577, opened 2024-05-24
fix(search:corpus): add status to corpus indices

kschelonka:
The service now emits events for draft collections, so we want to ensure we can differentiate between
published and draft collections when we return results.

Future enhancements might delete or discard this data, but since it's possible to transition from Published to Draft and there just isn't much data, it's not
expensive to just keep it around for now as we decide on exactly how we want to model.

Note that this field being passed was already covered in tests, it just wasn't included on the elasticsearch request. So it is covered in integration tests without any changes.

I have already updated dev and prod indices to accept this field.


nzeltzer (approved):
lgtm

PR #578, opened 2024-05-27
SE-3734 Putting client-api behind WAF

steveprokopienko:


bassrock (approved):
Looks good to me. Assuming all requests from our nat are always allowed

bassrock (approved):

kschelonka (approved):

PR #579, opened 2024-05-28
fix(sentry): updating sentry to 8.5.0

bassrock:
# Goal

Some of our lambdas are not emitting errors to sentry. This is because we did not init sentry in a variety of them. 

## Implementation Details

The scope here creeped and I not just added Sentry Init to those lambdas, but also upgraded Sentry to 8.5.0. 

Sentry 8.5.0 now includes native OpenTelemetry support which is what we use to do our performance tracing. I updated our tracing package to take advantage of this and if `addSentry` option is true, we will merge our AWS XRay Tracing with the [Sentry Open Telmetry tracing](https://docs.sentry.io/platforms/javascript/guides/express/migration/v7-to-v8/v8-opentelemetry/#accessing-the-opentelemetry-tracer). 

One thing I did is I had to create CustomSpanProcessors because we want to send a different volume of tracing events to Sentry then we do to AWS XRay (cost reasons) and OpenTelemetry does not support that out of the box. To support that I extended the respective span processors to first sample the trace based on the respective samplers before emitting the traces to their services.

https://mozilla-hub.atlassian.net/browse/POCKET-10151


docker-compose.yml
```diff
@@ -3,24 +3,14 @@ services:
...
-    command:
-      [
-        '--config',
-        '/etc/otelcol-contrib/config.yaml',
-        '--set=service.telemetry.logs.level=error',
-      ]
-    # If we want to set a custom config we would do it by mounting the file below.
-    # volumes:
-    #   - ./otel-collector-config.yaml:/etc/otelcol-contrib/config.yaml
+    image: grafana/otel-lgtm
```
bassrock, line 6: This image gives us a ui to view telemetry!

kschelonka (approved):

PR #582, opened 2024-05-29
fix(annotations-api): making no handler implemented a log instead of an error

bassrock:
## Goal

Fix a sentry bug that started popping up now that sentry is on the lambdas.

We should never throw on event handlers missing, unless the infrastructure will only ever send the appropriate handled events.

POCKET-10181


nzeltzer (approved):
lgtm

kschelonka (approved):

PR #584, opened 2024-05-29
chore(docs:search): add README for search subgraph

kschelonka:
Adding a README - no fancy graphics or anything, but just to get us started.


servers/user-list-search/README.md
```diff
@@ -0,0 +1,51 @@
+# Search API
```
bassrock, line 1: you are a good soul

bassrock (approved):

PR #615, opened 2024-06-26
fix(sentry): ensuring integrations are added to sentry

bassrock:
## Goal

It seems that a sentry update disabled the integrations we want by default if you don't pass in a trace sample rate option, even if its 0.

https://github.com/getsentry/sentry-javascript/blob/develop/packages/node/src/sdk/index.ts#L82-L84


kschelonka (approved):
Great find

PR #618, opened 2024-06-26
fix(search): update queries for os_2.13 upgrade

kschelonka:
[x] API tested in dev (could not test lambda integration because of calls to mysql db looking for premium user status)

Update localstack search cluster to be opensearch,
not elasticsearch. Note that the latest version we
could get locally is 2.11 but we are at 2.13 on AWS.

Remove 'type' (e.g. /_doc) from bulk query
endpoint.

Update score sort to explicitly include ordering,
rather than just passing the '_score' string.
Add a tiebreaker sort on 'item_id' for when _score
is identical among results.

Remove some unnecessary comparisons in test assertions
to clean up the code and streamline what we're checking.

[POCKET-10244]

[POCKET-10244]: https://mozilla-hub.atlassian.net/browse/POCKET-10244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/servers/user-list-search/src/datasource/elasticsearch/elasticsearchSearch.integration.ts
+++ b/servers/user-list-search/src/datasource/elasticsearch/elasticsearchSearch.integration.ts
@@ -25,7 +25,6 @@ describe('Elasticsearch Search Query', () => {
   beforeEach(async () => {
     await client.deleteByQuery({
       index: config.aws.elasticsearch.list.index,
-      type: config.aws.elasticsearch.list.type,
       body: {
         query: {
           match_all: {},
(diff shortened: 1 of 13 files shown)


servers/user-list-search/src/datasource/elasticsearch/elasticsearchSearch.integration.ts
```diff
@@ -317,7 +316,7 @@ describe('Elasticsearch Search Query', () => {
     expect(response.edges[1].node.savedItem.id).toBe(333);
   }, 10000);
 
-  it('should return search response with paginated fields first', async () => {
+  it('should return search response with paginated fields: first', async () => {
```
kschelonka, line 319: I examined the response from the search method and corrected this test. It's possible that the scoring algorithm is slightly different.

bassrock (approved):

bassrock (approved):

PR #624, opened 2024-07-01
fix(given_url): ensure we always return a given url

bassrock:
## Goal

POCKET-10284

Ensure we always return a given url if the item is pending in the V3 proxy layer.

I also added some list tests that seem useful before i discovered the issue was not in list api.


servers/v3-proxy-api/src/graph/get/toRest.ts
```diff
@@ -363,7 +363,7 @@ function ListItemTransformer<T extends SavedItemSimple>(
       return {
         ...baseFields,
         resolved_id: '',
-        given_url: '',
+        given_url: savedItem.url,
```
bassrock, line 366: This is the actual fix. Rest is all tests

kschelonka (approved):

PR #630, opened 2024-07-03
chore(corpus:search): denormalize ID for sorting

kschelonka:
Want to use the ApprovedItem ID for deterministic
sort response. To do it efficiently we should not
use the _id field but should denormalize the data
and copy it into a doc_values field.

**TODO**
- [x] Update dev mappings
- [x] Update prod mappings


nzeltzer (approved):
lgtm

PR #632, opened 2024-07-08
  feat(corpus:search): corpus search API - first cut implemenation

kschelonka:
First cut of corpus search API.

Calling out one change against the contract: Do not accept an array of query strings. It makes pagination and sorting almost impossible; instead let's just submit multiple queries or add additional "filters" if we want to do a title-only search and different text search, for example)

Does not yet include event emission (will follow up -- getting this out earlier will allow frontend clients to start testing)

I reorganized some files, so apologies for the import changes 🙇 

Another follow-up TODO is to make this accessible for logged-out users.

[POCKET-10137]

[POCKET-10137]: https://mozilla-hub.atlassian.net/browse/POCKET-10137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


servers/parser-graphql-wrapper/schema.graphql
```diff
@@ -540,3 +540,11 @@ type PocketShare @key(fields: "targetUrl") {
   targetUrl: ValidUrl!
   preview: PocketMetadata
 }
+
+"""A node in a CorpusSearchConnection result"""
+type CorpusSearchNode @key (fields: "url") {
+  """For federation only"""
+  url: Url! @inaccessible
```
bassrock, line 547: i like this idea, i see what you are doin

servers/user-list-search/src/datasource/clients/openSearch.ts
```diff
@@ -0,0 +1,16 @@
+import { Client } from '@opensearch-project/opensearch';
+import { config } from '../../config';
+
+const getHost = (host: string): string => {
```
bassrock, line 4: 😍

bassrock (approved):

PR #639, opened 2024-07-11
feat(auth:search): use directive-based authZ rules for search

kschelonka:
Use directives for query authorization. This allows us to remove the context's assertion for logged-in users on user-list-search without doing a more significant refactor of code flow.

Testing:
- [x] Dev deployment
- [ ] ~V3 API Proxy requests to dev for affected search endpoints~ Can't test this with our current infrastructure -- will do a staged deployment where we temporarily disable the proxy deployment, then test with only our test users enabled; waiting on an AWS incident which is preventing us from authorizing into unleash
- [x] Manual dev QA (premium, non-premium, anonymous) for all affected search endpoints

Changes:
* Add two rhai scripts for
    * Converting roles array into space-separated scope claim as per [OAuth2 RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3)
    * Extracting key-value claims (e.g. `is_application_native:true`) for `@policy` directive
* Remove requirement for non-anonymous user on UserListSearch's GraphQL Context

Did not:
* Remove embedded auth checks in code flow (but I probably could -- worst case it's no-op/double checking)


bassrock (approved):
easy peasy

PR #651, opened 2024-07-17
fix(since): truncate decimal based since values

bassrock:
## Goal

Fix some old iOS client 400s causing noise by truncated their decimal based since values

https://mozilla-hub.atlassian.net/browse/POCKET-10347?atlOrigin=eyJpIjoiNTRlYzdiNTQzNjk3NDdlMGIwZDA4OGI4NTFmZTM2NDciLCJwIjoiamlyYS1zbGFjay1pbnQifQ

### DIFF
--- a/servers/v3-proxy-api/src/routes/validations/GetSchema.ts
+++ b/servers/v3-proxy-api/src/routes/validations/GetSchema.ts
@@ -108,11 +108,6 @@ export const V3GetSchema: Schema = {
   },
   since: {
     optional: true,
-    isInt: {
-      options: {
-        min: 0,
-      },
-    },
     toInt: true,
     customSanitizer: {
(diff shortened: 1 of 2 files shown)


servers/v3-proxy-api/src/routes/validations/GetSchema.ts
```diff
@@ -108,11 +108,6 @@ export const V3GetSchema: Schema = {
   },
   since: {
     optional: true,
-    isInt: {
-      options: {
-        min: 0,
```
kschelonka, line 113: removing isInt makes sense, but can you please add a validator method to check it's not < 0 (otherwise may throw an unhandled serialization error)

kschelonka (approved):

PR #652, opened 2024-07-17
feat(snowplow): adding in snowplow session & userid to list-api

bassrock:
## Goal

Transmit the snowplow session value and user id that should be set

https://mozilla-hub.atlassian.net/browse/POCKET-10342

Requires snowplow session to be added to the router headers


kschelonka (approved):

kschelonka (changes requested):
Can we change this to the name of the claim in the JWT so that I don't have to write a custom router forwarder

snowplow_domain_session_id

servers/client-api/config/router.yaml
```diff
@@ -52,6 +52,9 @@ headers:
           named: 'X-Amzn-Trace-Id'
       - propagate:
           named: 'transfersub'
+      - propagate:
```
kschelonka, line 55: @bassrock  please confirm, value from https://github.com/Pocket/Web/blob/ea9b2402eebbd266722954561ef47ceda61973bd/src/ClientAPI/Client.php#L219

kschelonka (approved):

servers/client-api/config/router.yaml
kschelonka: confirmed via slack

PR #660, opened 2024-07-18
fix(deps): updating dev deps

bassrock:
## Goal

Update a bunch of development dependencies at once.


servers/list-api/src/dataService/savedItemsService.ts
```diff
@@ -293,7 +293,7 @@ export class SavedItemDataService {
   ) {
     const input = rows.map((row) =>
       Object.keys(row).reduce((obj, key) => {
-        if (row[key] instanceof Date && isNaN(row[key])) {
+        if (isNaN(row[key]) && row[key] instanceof Date) {
```
bassrock, line 296: new version of typescript caught this!

kschelonka: maybe it's the diff but what is the change here? the ordering of the statements?

kschelonka (approved):

PR #663, opened 2024-07-18
fix(send): ensuring a send add resposne returns items not nested

bassrock:
## Goal

It seems the Chrome extension expects the item in the response array not nested like:

```json
action_results: [{status: 1, item: {}}]
```

but rather:

```json
action_results: [{status: 1, ...itemFieldsHere}]
```

I turned off the v3Proxy for send and recorded requests that confirmed this.

https://mozilla-hub.atlassian.net/browse/POCKET-10355


servers/v3-proxy-api/src/routes/v3Send.integration.ts
```diff
@@ -555,7 +555,21 @@ describe('v3/send', () => {
               input: {
                 action: 'tags_replace',
                 item_id: '12345',
-                tags: 'supplemental',
+                tags: ['supplemental'],
```
bassrock, line 558: just adding some test cases i saw chrome sends

kschelonka (approved):

PR #666, opened 2024-07-19
feat(search events): implement search event emitter and consumer

kschelonka:
Do not merge until next week.

Emit search events to the event bus when corpus searches are performed.

Stand up infrastructure for the Event Bridge integration, with SNS topic, listeners, and consumer (via shared-snowpow-consumer).

Uses types generated from snowplow in both the emitter and consumer to keep schemas in sync across services.


nzeltzer (approved):
lgtm

PR #671, opened 2024-08-07
fix(v3proxy:extensions): flatten /send response for extensions

kschelonka:
Note that I have some extension-specific code checks which I ended up not using in this PR, but they will be used in subsequent PRs (due to changing `/v3/get` response for extensions).

This is a follow up from #663, which we reverted due to reports from 3rd party API users.

I am reasonably certain that _only_ the `/v3/send` response for the `add` action is flattened, and not the response for `/v3/add` itself.

Investigation and testing:
* When #663 was merged, I received bug reports from 3rd party API users about the unnesting
* Our [developer documentation](https://getpocket.com/developer/docs/v3/add) indicates that `/v3/add` response data is nested under the `item` key
* Our native apps and extensions return nested `/v3/add` results (tested with ios and chrome extension consumer keys)
* The response type is **not** documented for `/v3/send`, but requests made with both ios and chrome extension consumer keys returned the flattened (unnested) result
* There is special handling for native apps and extensions, but it doesn't appear to be called in these code paths (discovered elsewhere, to be continued)
* Android handles errors and makes a fresh `fetch` request, so it was not materially impacted by either of the `add` or `send` response shapes 

[POCKET-19379]

### DIFF
--- a/servers/v3-proxy-api/src/graph/graphQLClient.ts
+++ b/servers/v3-proxy-api/src/graph/graphQLClient.ts
@@ -157,7 +157,7 @@ export class GraphQLClientFactory {
             !(
               error.path &&
               Array.isArray(error.path) &&
-              error.path.indexOf('recentSearches') &&
+              error.path.indexOf('recentSearches') >= 0 &&
               error.extensions.code === 'FORBIDDEN'
             ),
         );
--- a/servers/v3-proxy-api/src/config/index.ts
+++ b/servers/v3-proxy-api/src/config/index.ts
@@ -6,6 +6,12 @@ export default {
     defaultMaxAge: 86400,
     port: 4030,
   },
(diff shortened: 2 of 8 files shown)


servers/v3-proxy-api/src/graph/graphQLClient.ts
```diff
@@ -157,7 +157,7 @@ export class GraphQLClientFactory {
             !(
               error.path &&
               Array.isArray(error.path) &&
-              error.path.indexOf('recentSearches') &&
+              error.path.indexOf('recentSearches') >= 0 &&
```
kschelonka, line 160: small bugfix, I don't think `recentSearches` is ever possible to be in the leftmost (zero-indexed) path but let's fix it anyway

collectedmind (approved):

PR #697, opened 2024-08-24
feat(preview): adding standardized preview data cards

bassrock:
## Goal

Standardize all display logic across Pocket using the Preview Entity.

Requires: https://github.com/Pocket/content-monorepo/pull/210

## Implementation Details

I chose to specifically use GraphQL to do our intersystem communication so that we could focus on merging logic. For now this works pretty well and the router handles all flattening in our resolvers

https://mozilla-hub.atlassian.net/browse/POCKET-10474


servers/parser-graphql-wrapper/schema-admin.graphql
```diff
@@ -0,0 +1,556 @@
+extend schema
```
bassrock, line 1: I had to add a seperate schema for admin because federation gets mad since our admin api does not implement some of the shared objects

servers/client-api/config/supergraph.yaml
```diff
@@ -1,10 +1,76 @@
...
+  image-api:
+    routing_url: https://image-api.readitlater.com
+    schema:
+      subgraph_url: https://image-api.readitlater.com
+  list-api:
+    routing_url: https://list-api.readitlater.com
+    schema:
+      subgraph_url: https://list-api.readitlater.com
+  parser:
+    #routing_url: https://parser-graphql-wrapper.readitlater.com
```
kschelonka, line 24: reason this one is commented out?

servers/client-api/config/supergraph.yaml
```diff
@@ -1,10 +1,76 @@
...
+  user-list-search:
+    routing_url: https://user-list-search.readitlater.com/graphql
+    schema:
+      subgraph_url: https://user-list-search.readitlater.com/graphql
+
+  # # Content Monorepo
+  collection:
+    routing_url: https://collection-api.readitlater.com
+    # schema:
+    #   subgraph_url: https://collection-api.readitlater.com
```
kschelonka, line 51: are these below commented out due to no introspection?

servers/client-api/config/supergraph.yaml
bassrock: Was running it locally against my local parser

servers/client-api/config/supergraph.yaml
bassrock: yeppp

collectedmind (approved):

PR #720, opened 2024-09-06
feature(search): corpus search infrastructure, backfill, and ingest pipeline

kschelonka:
Infrastructure for corpus search, with ingest and backfill 

TODOs:
- [x] e2e manual testing in dev of corpus ingest and backfill process
- [x] Pre-deploy corpus search and sagemaker infrastructure (for below step)
- [x] Update lambdas with environment variables  - search endpoint and sagemaker endpoint
- [x] Add indices to prod corpus search cluster

Out of scope:
- Keyword search on corpus search cluster (cut over after backfill is done)
- Semantic search on user-list-search api (next up)


kschelonka:
Added the Sentry DSN value to the prod account. I'm looking into the corpus search cluster to see if there is any way I can get around having to destroy it.

PR #740, opened 2024-09-17
docs(description): making id more clear in image api

bassrock:
## Goal

Make the description a bit better for @collectedmind  ;) 

_all good jest_

### DIFF
--- a/servers/image-api/schema.graphql
+++ b/servers/image-api/schema.graphql
@@ -35,7 +35,7 @@ enum ImageFileType {
 Set of parameters that will be used to change an image
 """
 input CachedImageInput {
-  "ID that will be added to the generated response so you can find it you silly face"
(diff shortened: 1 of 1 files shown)


collectedmind:
@collectedmind That F'n guy ... some people just can't be trusted.

servers/image-api/schema.graphql
```diff
@@ -35,7 +35,7 @@ enum ImageFileType {
 Set of parameters that will be used to change an image
 """
 input CachedImageInput {
-  "Id of the image in the returned result set"
+  "ID that will be added to the generated response so you can find it you silly face"
```
collectedmind, line 38: Perhaps:

ID that will be added to the generated response object so you can find it.  Can be any string that you will use to delineate it when consumed

collectedmind (approved):

PR #743, opened 2024-09-18
feat(variants): support the new variants in unleash

bassrock:
## Goal

Unleash 6.0 is moving to strategy based variants instead of toggle based, this change supports that on the graphql endpoint and is backwards compatible.

https://mozilla-hub.atlassian.net/browse/POCKET-10611


servers/feature-flags/src/graphql/resolvers.ts
```diff
@@ -60,28 +60,20 @@ const convertToUnleashAssignment = (
...
   return toggles.map((toggle: FeatureInterface) => {
-    // Unleash in a recent update returns null or an empty array for variants in the API response
-    // We need to check for both
-    if (
-      !toggle.variants ||
-      (toggle.variants instanceof Array && toggle.variants.length === 0)
-    ) {
+    const variant = instance.forceGetVariant(toggle.name, context);
+
+    if (!variant || variant.name == 'disabled') {
```
kschelonka, line 65: I'm a little confused about this. can you explain why 'disabled' name is included in this?

kschelonka: I guess if it's not making an external request for isEnabled then there's not really an issue -- would just return false

kschelonka (approved):

servers/feature-flags/src/graphql/resolvers.ts
bassrock: @kschelonka yea.. its how unleash is doing it now. https://docs.getunleash.io/reference/strategy-variants#the-disabled-variant

Basically, if your flag is turned on, but has no variants, the variant you get is "disabled". In this scernario, we just return whether or not the flag is on for a user.

PR #748, opened 2024-09-19
Revert "Revert "feat(variants): support the new variants in unleash""

bassrock:
Reverts Pocket/pocket-monorepo#747

### DIFF
--- a/servers/feature-flags/src/graphql/resolvers.ts
+++ b/servers/feature-flags/src/graphql/resolvers.ts
@@ -73,7 +73,7 @@ const convertToUnleashAssignment = (
       name: toggle.name,
       assigned: variant.enabled,
       variant: variant.name,
-      payload: variant.payload.value,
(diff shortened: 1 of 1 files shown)


servers/feature-flags/src/graphql/resolvers.ts
```diff
@@ -60,28 +60,20 @@ const convertToUnleashAssignment = (
...
-    );
-
     return {
       name: toggle.name,
-      assigned: enabled,
-      variant: name,
-      payload: payload?.value,
+      assigned: variant.enabled,
+      variant: variant.name,
+      payload: variant.payload?.value,
```
bassrock, line 76: ? was the change....

kschelonka: ah it was nullable

kschelonka (approved):

PR #749, opened 2024-09-19
fix(strictNull): adding strict null checks for packages

bassrock:
## Goal

Enable strict null checks for all our packages


packages/apollo-utils/src/scalars/ValidUrlScalar.ts
```diff
@@ -10,16 +10,17 @@ export const GraphQLValidUrl = /*#__PURE__*/ new GraphQLScalarType({
     'A field whose value conforms to the standard URL format as specified in RFC3986: https://www.ietf.org/rfc/rfc3986.txt.',
 
   serialize(value) {
-    if (value === null) {
+    if (value === null || value === undefined) {
```
kschelonka, line 13:
```suggestion
    if (value == null) {
```

packages/apollo-utils/src/scalars/ValidUrlScalar.ts
```diff
@@ -10,16 +10,17 @@ export const GraphQLValidUrl = /*#__PURE__*/ new GraphQLScalarType({
...
       return value;
     }
 
     return new URL(value.toString()).toString();
   },
 
-  parseValue: (value) => (value === null ? value : new URL(value.toString())),
+  parseValue: (value) =>
+    value === null || value === undefined ? value : new URL(value.toString()),
```
kschelonka, line 21:
```suggestion
    value == null ? value : new URL(value.toString()),
```

packages/apollo-utils/src/scalars/ValidUrlScalar.ts
```diff
@@ -30,7 +31,7 @@ export const GraphQLValidUrl = /*#__PURE__*/ new GraphQLScalarType({
     }
 
     if (ast.value === null) {
```
kschelonka, line 33: this should be `==` as well

packages/apollo-utils/src/sentry/apolloSentryPlugin.ts
```diff
@@ -52,7 +52,7 @@ export const sentryPlugin: ApolloServerPlugin<BaseContext> = {
         for (const err of ctx.errors) {
           // Only report internal server errors,
           // errors extending ApolloError should be user-facing
-          if (NO_REPORT_ERRORS.has(err.extensions?.code?.toString())) {
+          if (NO_REPORT_ERRORS.has(err.extensions?.code?.toString() ?? '')) {
```
kschelonka, line 55: Rather than looking for an empty string in the set, I just wouldn't do the lookup if you don't have that key

e.g.
```
const errCode = err.extensions?.code?.toString()
if (errCode && NO_REPORT_ERRORS.has(errCode) {...}
```

packages/ts-logger/src/morgan.ts
```diff
@@ -9,28 +9,28 @@ const defaultLogger: Logger = createLogger();
 
 // if userid header used, inject it into HTTP request logs
 token('req-userid', function (req, _res) {
-  const userid: string | string[] = req.headers.userid || null;
+  const userid: string | string[] | null = req.headers.userid || null;
   if (userid === null) {
```
kschelonka, line 13: this should be `==`

packages/ts-logger/src/morgan.ts
```diff
@@ -9,28 +9,28 @@ const defaultLogger: Logger = createLogger();
...
   }
   return userid.toString();
 });
 
 // if graph requestId header used, inject it into HTTP request logs
 token('req-requestId', function (req, _res) {
-  const requestId: string | string[] =
+  const requestId: string | string[] | null =
     req.headers['x-graph-request-id'] || null;
   if (requestId === null) {
```
kschelonka, line 23: `==`

kschelonka (approved):

PR #751, opened 2024-09-20
feat(corpus search): deduplicate syndicated articles

kschelonka:
Remove the original article from the corpus
when a syndicated article is added. Assumes
that the original article is added to the
corpus first -- maybe there are some unlikely
edge cases which will result in duplicates, but
the tradeoff is worth it so that we don't have to
check every new write.

Choosing to delete rather than hide because the
syndicated article, if it expires or is removed,
will simply redirect to the original page.

[POCKET-10555]

[POCKET-10555]: https://mozilla-hub.atlassian.net/browse/POCKET-10555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lambdas/user-list-search-corpus-indexing/src/index.ts
```diff
@@ -47,21 +47,36 @@ export const handler = Sentry.wrapHandler(processor);
...
+  const validItems: ValidatedEventPayload[] = records.filter((item) =>
+    validLanguage(item),
+  );
+  // Delete syndication duplicates
+  // This assumes that for a given syndicated article, the
+  // original article existed in the corpus *before* it was
+  // was syndicated. We do not check for a possible syndication
+  // duplicate whenever a new article is added, because that would
+  // be resource-intensive and generally doesn't make sense with the
+  // curation workflow. We'll take the very small risk of duplicates
```
bassrock, line 73: 👍

bassrock (approved):

PR #752, opened 2024-09-20
fix(strictNull): adding strict null check to lambdas

bassrock:
## Goal

Add strict null checks to the lambdas

### DIFF
--- a/lambdas/account-data-deleter-batch-delete/src/config.ts
+++ b/lambdas/account-data-deleter-batch-delete/src/config.ts
@@ -1,10 +1,10 @@
 const shortName = 'ADD';
 
 const awsEnvironments = ['production', 'development'];
-let localAwsEndpoint: string | undefined = undefined;
-if (!awsEnvironments.includes(process.env.NODE_ENV ?? '')) {
-  localAwsEndpoint = process.env.AWS_ENDPOINT || 'http://localhost:4566';
-}
+const localAwsEndpoint =
+  process.env.NODE_ENV && !awsEnvironments.includes(process.env.NODE_ENV)
+    ? process.env.AWS_ENDPOINT || 'http://localhost:4566'
+    : undefined;
 
 export const config = {
   app: {
--- a/lambdas/account-data-deleter-batch-delete/src/config.ts
+++ b/lambdas/account-data-deleter-batch-delete/src/config.ts
@@ -1,8 +1,8 @@
 const shortName = 'ADD';
 
 const awsEnvironments = ['production', 'development'];
-let localAwsEndpoint = undefined;
-if (!awsEnvironments.includes(process.env.NODE_ENV)) {
+let localAwsEndpoint: string | undefined = undefined;
+if (!awsEnvironments.includes(process.env.NODE_ENV ?? '')) {
   localAwsEndpoint = process.env.AWS_ENDPOINT || 'http://localhost:4566';
 }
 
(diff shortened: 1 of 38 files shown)


lambdas/user-list-search-corpus-indexing/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "tsconfig/lambda.json",
   "compilerOptions": {
     "outDir": "dist",
-    "rootDir": "src"
+    "rootDir": "src",
+    "strictNullChecks": false
```
bassrock, line 6: Turned off to not conflict with https://github.com/Pocket/pocket-monorepo/pull/751

lambdas/account-data-deleter-batch-delete/src/config.ts
```diff
@@ -1,8 +1,8 @@
 const shortName = 'ADD';
 
 const awsEnvironments = ['production', 'development'];
-let localAwsEndpoint = undefined;
-if (!awsEnvironments.includes(process.env.NODE_ENV)) {
+let localAwsEndpoint: string | undefined = undefined;
```
kschelonka, line 4: don't need to change this, but I'd prefer a ternary statement vs. initialization and conditional assignment for our code generally

lambdas/transactional-emails/src/schemas/premiumPurchaseSchema/premiumPurchaseEvent.ts
```diff
@@ -3,7 +3,7 @@ import { User } from './user';
 
 export class PremiumPurchaseEvent {
   'purchase': Purchase;
-  'user': User;
+  'user': User | null;
```
kschelonka, line 6: how is this nullable?

lambdas/transactional-emails/src/schemas/premiumPurchaseSchema/purchase.ts
```diff
@@ -5,7 +5,7 @@ export class Purchase {
   'isTrial': boolean;
   'planInterval': string;
   'planType': string;
-  'receiptId': string;
+  'receiptId': string | null;
```
kschelonka, line 8: how is this nullable?

lambdas/account-data-deleter-batch-delete/src/config.ts
bassrock: @kschelonka `const localAwsEndpoint =
  process.env.NODE_ENV && !awsEnvironments.includes(process.env.NODE_ENV)
    ? process.env.AWS_ENDPOINT || 'http://localhost:4566'
    : undefined;`

lambdas/user-list-search-corpus-indexing/tsconfig.json
kschelonka: #751 is merged

lambdas/user-list-search-corpus-parser-hydration/src/index.ts
```diff
@@ -121,7 +122,11 @@ export function unwrapPayloads(payload: EventPayload): BulkRequestMeta[] {
       messageId: payload.messageId,
     };
     return [parent, ...stories];
-  } else {
+  } else if (
```
kschelonka, line 125: noting that this is already evaluated before passed to the function; will update in follow-up

kschelonka (approved):

PR #753, opened 2024-09-20
fix(corpus search): do not index embeddings for articles without excerpts

kschelonka:
With the exception of collections which might not have excerpts.

This is because the existence of an excerpt on a Corpus item is a good proxy indicator of quality/review. Items without excerpts tend to have poor metadata (e.g. a url for a title).

[POCKET-10583]

[POCKET-10583]: https://mozilla-hub.atlassian.net/browse/POCKET-10583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lambdas/user-list-search-corpus-parser-hydration/src/index.ts
+++ b/lambdas/user-list-search-corpus-parser-hydration/src/index.ts
@@ -7,16 +7,15 @@ Sentry.init({
 });
 import type { SQSBatchResponse, SQSEvent } from 'aws-lambda';
 import {
-  EventPayload,
   validDetailTypes,
   BulkRequestMeta,
   BulkRequestPayload,
+  ValidLangEventPayload,
 } from './types';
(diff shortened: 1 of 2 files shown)


lambdas/user-list-search-corpus-parser-hydration/src/index.ts
```diff
@@ -120,13 +130,10 @@ export function unwrapPayloads(payload: EventPayload): BulkRequestMeta[] {
       title: collection.title,
       excerpt: collection.excerpt,
       messageId: payload.messageId,
+      isCollection: true,
     };
     return [parent, ...stories];
-  } else if (
```
kschelonka, line 125: null fallbacks are implemented downstream so we don't need this check

bassrock (approved):

PR #755, opened 2024-09-23
fix(strictNull): adding in strict null checks for Braze Content Proxy

bassrock:
## Goal

Add in strict null checks for braze content proxy. https://mozilla-hub.atlassian.net/browse/POCKET-10095


servers/braze-content-proxy/src/routes/digest.ts
```diff
@@ -40,17 +40,20 @@ function transformToBrazePayload(
...
+          title: item.preview?.title ?? null,
+          imageUrl: item.preview?.image?.cachedImages?.at(0)?.url ?? null,
+          url: item.preview?.url ?? null,
         };
       }
-    });
+      return null;
+    })
+    // Added due to strict null checks and types above not inferring
+    .filter((item) => item != null) as BrazeSavedItem[];
```
kschelonka, line 56: it's because your map above can return null

bassrock: @kschelonka found a better way

servers/braze-content-proxy/src/routes/digest.ts
```diff
@@ -40,16 +40,17 @@ function transformToBrazePayload(
...
     .map((edge) => {
-      const item = edge.node.item;
-      if (item.__typename === 'Item') {
-        return {
-          title: item.preview.title,
-          imageUrl: item.preview.image?.cachedImages[0].url,
-          url: item.preview.url,
-        };
-      }
+      const item = edge?.node?.item as Item; // force cast to Item because map can't infer type from the above filter.
```
kschelonka, line 48: you can use a type predicate in the filter function if you want. it's not too different to what you're doing here but just moves the hint up

```
  .filter(
      (edge): edge is Item =>
        edge?.node?.item != null && edge?.node?.item.__typename === 'Item',
    )
    ```

kschelonka (dismissed):

kschelonka (approved):

PR #757, opened 2024-09-23
fix(strictNull): adding strict null for account data deleter

bassrock:
## Goal

Add in strict null for account data deleter

### DIFF
--- a/servers/account-data-deleter/src/config/tables.ts
+++ b/servers/account-data-deleter/src/config/tables.ts
@@ -18,7 +18,10 @@ const nonUserTables = [
 //excludes any table that is handled by list-api,
 //annotations-api, and user-api
 //delete by `user_id`
-const tablesWithUserId = {
+const tablesWithUserId: {
+  tablesWithSensitivePii: string[];
+  tablesWithUserIdAlone: string[];
+} = {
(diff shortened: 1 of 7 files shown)


servers/account-data-deleter/src/config/tables.ts
```diff
@@ -193,9 +196,10 @@ const tablesWithUserId = {
   ],
 };
 
-const userTables = []
-  .concat(tablesWithUserId.tablesWithSensitivePii)
-  .concat(tablesWithUserId.tablesWithUserIdAlone);
+const userTables: string[] = [
+  ...tablesWithUserId.tablesWithSensitivePii,
```
bassrock, line 200: 90% sure this is doing the same thing.. and it doesnt give type errors.

kschelonka: oh I bet the original type error is because for some reason this was constructed with `[].concat(<some_list>).concat(<another_list>) vs. `some_list.concat(<another_list>)`

kschelonka: can't infer element type on an empty array

kschelonka (approved):

PR #762, opened 2024-09-23
fix(list-api): seperate out read and write

bassrock:
## Goal

Some queries can cause writes to the db, so instead both read & write clients should pass through to services, and they should choose the right client to use for the query.


servers/list-api/src/test/graphql/mutations/savedItemById/createSavedItemTagsMutation.integration.ts
```diff
@@ -207,7 +207,7 @@ describe('createSavedItemTags mutation', function () {
     expect(eventData[2]).toContainAllValues(['tofino', 'victoria']);
   });
 
-  it('mutations resolver chains should call only writeClient()', async () => {
+  it('mutations resolver chains call both clients', async () => {
```
kschelonka, line 210: probably this test can be deleted as it no longer has the same purpose

kschelonka (approved):

PR #793, opened 2024-10-02
fix(v3proxy): state default to unread

kschelonka:
update the default state to match our 3rd party dev docs

cc @marcin-kozinski to confirm this change in default is OK with android requests


bassrock (approved):

kschelonka:
This doesn't comply with Android expectations so we must update our 3rd party api docs instead.

PR #794, opened 2024-10-03
feat(otel): adding in basic tracing via otel

bassrock:
# Goal

Setup APM via Otel to gain insights into system performance.

Note: In dev testing this increased memory by 2%. The OTEL collector also has memory protection in place to drop spans if it gets too big.

We also have [Logs](https://console.cloud.google.com/logs/query;query=labels.%22service.name%22%3D%22v3-api-proxy%22;cursorTimestamp=2024-10-07T14:20:06.351Z;duration=PT1H?project=moz-fx-pocket-prod-61fb) in GCP.

## Moving forward

Based on our system and how tracing works, I believe our future iterations should focus mainly on Sampling from within Client API and allow the downstream services to sample based on [their parent trace](https://github.com/Pocket/pocket-monorepo/pull/794/files#diff-aa8814b8d76fb52a8b215b8e5ae4e4e39419143ab71976ae34eb07fc35525316R170) OR if they have an error only. 

Right now having the ability to sample each service based on a % is useful for debugging as we roll this out, but for day to day observability it is more helpful to only capture full system traces, which we will only get if Client API decides to sample a trace and all downstream calls.  [Follow up ticket here.](https://mozilla-hub.atlassian.net/browse/POCKET-10675)

Local Screenshot
image

GCP Screenshot

[Example](https://console.cloud.google.com/traces/list?referrer=search&project=moz-fx-pocket-prod-61fb&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22P4D%22,%22customValue%22:null))&tid=abe9bd3179391087aa7216f59b5dfe8e)

Screenshot 2024-10-07 at 6 44 00 AM


packages/tracing/src/tracing.ts
```diff
@@ -128,50 +115,45 @@ export async function nodeSDKBuilder(config: TracingConfig) {
...
-    new HttpInstrumentation({
-      ignoreIncomingPaths: ['/.well-known/apollo/server-health'],
+    getNodeAutoInstrumentations({
+      '@opentelemetry/instrumentation-fs': {
+        enabled: false,
+        requireParentSpan: true,
+      },
+      '@opentelemetry/instrumentation-undici': {
+        headersToSpanAttributes: {
+          requestHeaders: [
```
bassrock, line 144: These are headers that we add to our trace data, we could expand this in the future. This should be enough to correlate with sentry

kschelonka: does applicationname map to consumer key?

kschelonka (approved):

PR #826, opened 2024-10-14
fix(bools): allowing true/false in a post body for v3

bassrock:
## Goal

I noticed https://pocket.sentry.io/issues/5623927204/events/93e309e8c6a54ef5bc55fb0b1102beb5/?project=4504658424102912 errors are all using booleans in the post bodies. 

This adds back that support that Web Repo supported

### DIFF
--- a/servers/v3-proxy-api/src/routes/validations/FetchSchema.ts
+++ b/servers/v3-proxy-api/src/routes/validations/FetchSchema.ts
@@ -102,10 +102,11 @@ export const V3FetchSchema: Schema = {
       options: '0',
     },
     isIn: {
-      options: [['0', '1', true, false]],
+      options: [['0', '1', true, false, 'true', 'false']],
     },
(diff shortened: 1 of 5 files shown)


kschelonka (dismissed):
good find. we all love undocumented schemas :sob:

servers/v3-proxy-api/src/routes/validations/FetchSchema.ts
```diff
@@ -102,43 +102,43 @@ export const V3FetchSchema: Schema = {
       options: '0',
     },
     isIn: {
-      options: [['0', '1']],
+      options: [['0', '1', true, false]],
```
kschelonka, line 105: probably need to stringify these

bassrock: updated

kschelonka (approved):

PR #833, opened 2024-10-15
fix(logging): adding in logging to v3 errors

bassrock:
# Goal

Trace down the root of v3 errors that are showing no errors but appear in sentry.


kschelonka (approved):
the breadcrumb seems kind of unnecessary here but I won't block on this

PR #850, opened 2024-10-16
feat(tracing): adding tracing to annotations-api

bassrock:
# Goal

Add tracing/log exporting to annotations api


servers/v3-proxy-api/src/config/index.ts
```diff
@@ -22,7 +22,6 @@ export default {
     dsn: process.env.SENTRY_DSN || '',
     release: process.env.GIT_SHA || 'local',
     environment: process.env.NODE_ENV || 'development',
-    samplerFlag: 'perm.backend.sentry-trace-sampler-rate',
```
bassrock, line 25: no longer needed

kschelonka (approved):

PR #852, opened 2024-10-16
fix(logging): updating logs to not use the reserved "data" key

bassrock:
# Goal

Update our logging statements to not use `data` since our logger deletes it due to our graphql resolvers auto-filling that key with a ton of uneedded data. 

Long term we should do more permisive filtering of the data key instead of just deleting it.


servers/account-data-deleter/src/dataService/accountDeleteDataService.ts
```diff
@@ -51,7 +51,7 @@ export class AccountDeleteDataService {
 
     serverLogger.error({
       message: 'getIndexColumnForTables: lookup failed.',
-      data: {
+      errorData: {
```
bassrock, line 54: Seeing this error in GCP, looking to resolve it.

kschelonka (approved):

PR #864, opened 2024-10-17
fix(client-api): adding in router telemetry metrics

bassrock:
# Goal

Enable Apollo Studio metrics.

Since Apollo Router does not have any OTEL resource detectors this uses a shell script to grab the data from the AWS ECS Endpoint and then we set env variables to the values we need, which are then imported into the config file.


kschelonka (approved):

servers/client-api/Dockerfile
```diff
@@ -2,12 +2,21 @@ FROM ghcr.io/apollographql/router:v1.56.0@sha256:b6a69cb14c73d890e32a5ab3c47508d
...
 
 COPY ./config/router.yaml /config/router.yaml
 
 COPY ./rhai ./rhai
 
-CMD ["--config", "/config/router.yaml"]
+# Set the entrypoint to the script
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+
+CMD ["/dist/router_wrapper.sh", "--config", "/config/router.yaml"]
```
bassrock, line 22: this is the original entrypoint script

servers/client-api/entrypoint.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/sh
+
+# Fetch ECS metadata
+ECS_METADATA=$(curl -s ${ECS_CONTAINER_METADATA_URI_V4}/task)
```
bassrock, line 4: Provided by AWS.

kschelonka (approved):

PR #871, opened 2024-10-21
fix(charset): rewriting the charset for bad utf8 values

bassrock:
# Goal

Our old API system allowed for invalid UTF8 charset values.

To account for this until clients can update their requests, this will rewrite UTF8 to UTF-8 as the first thing within our request middleware.


servers/v3-proxy-api/src/routes/v3Get.integration.ts
```diff
@@ -1719,4 +1719,66 @@ describe('v3Get', () => {
...
+        .send({
+          consumer_key: 'test',
+          access_token: 'test',
+          since: 1712766000,
+          detailType: 'complete',
+        })
+        .set('Content-Type', 'application/json; charset=UTF8');
+      expect(response.body).toEqual(expectedGetComplete);
+      expect(requestSpy).toHaveBeenCalledTimes(1);
+      expect(requestSpy.mock.calls[0][3]).toMatchObject({
```
kschelonka, line 1742: this checks that the request is made (makes through the try/catch) but not that the utf8 header is rewritten. Can you check the headers on the request if they're forwarded?

bassrock: @kschelonka  Content-Type gets deleted on forwarding, this is mainly on express accepting the value.

PR #874, opened 2024-10-22
fix(optional): making isPremium optional

bassrock:
# Goal

isPremium can also be optional coming into this Lambda body.


lambdas/account-data-deleter-events/src/handlers/accountDelete/index.ts
```diff
@@ -8,7 +8,7 @@ import { AggregateError } from '../../errors/AggregateError';
 
 type AccountDeleteBody = {
   userId: string;
-  email: string;
+  email?: string;
   isPremium: boolean;
```
kschelonka, line 12:
```suggestion
  isPremium?: boolean;
```

kschelonka (approved):

PR #875, opened 2024-10-22
fix(router): ensuring we append v1/metrics

bassrock:
# Goal

Append v1/metrics to the routers otel sender, and add in some local testing scripts.

### DIFF
--- a/servers/client-api/config/router.yaml
+++ b/servers/client-api/config/router.yaml
@@ -204,7 +204,7 @@ telemetry:
                   - 1000000
       otlp:
         enabled: true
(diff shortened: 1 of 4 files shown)


servers/client-api/config/router.yaml
```diff
@@ -204,7 +204,7 @@ telemetry:
                   - 1000000
       otlp:
         enabled: true
-        endpoint: '${env.OTLP_COLLECTOR_URL:-http://localhost:4318}'
+        endpoint: '${env.OTLP_COLLECTOR_URL:-http://localhost:4318}/v1/metrics'
```
bassrock, line 207: The router doesn't append v1/metrics

kschelonka (approved):

PR #907, opened 2024-11-01
feat(import): omnivore list import and presigned url generation

kschelonka:
The list-api ingest part of the upload

I tested in dev and was able to upload using the signed url
Files flowed through the pipeline, but I hit a snag because the parser wasn't returning resolved_id for some items it hadn't previously parsed

Usage:

web client requests a presigned url and uploads the zipfile to s3 (with reasonable size limits)
Requires native pocket client and logged-in user. If there hasn't already been an upload on that calendar day for that user, an upload link is provided. Resets the next calendar day (UTC)

example mutation:

```
mutation importUploadUrl($importType: ImportType!) {
  importUploadUrl(importType: $importType) {
    ... on PreSignedUrl {
      url
    }
    ... on ImportLimited {
      __typename
    }
  }
}
```

Out of scope/left todo:
* backend events
* giving a better time estimate on the signedUrl rate limit response
* making packages out of reused methods


infrastructure/list-api/src/main.ts
```diff
@@ -210,14 +211,14 @@ class ListAPI extends TerraformStack {
               name: 'SQS_PUBLISHER_DATA_QUEUE_URL',
               value: `https://sqs.${region.name}.amazonaws.com/${caller.accountId}/${config.envVars.sqsPublisherDataQueueName}`,
             },
-            {
```
bassrock, line 213: Asssuming this was just old infra?

servers/list-api/src/background/batchImportHandler.ts
```diff
@@ -0,0 +1,229 @@
...
+import { getClient } from '../featureFlags';
+import { SavedItemImportInput } from '../types';
+import gql from 'graphql-tag';
+import { print } from 'graphql';
+
+/**
+ * A slightly silly config to override a value conditionally
+ * with a feature flag
+ */
+class ImportHandlerQueueConfig implements QueueConfig {
```
bassrock, line 17: (out of scope for this pr): Def see a use case for making an SQS polling package or updating to use https://github.com/bbc/sqs-consumer across the board in a package or something. This is probably now number 4? of a poller we have?

servers/list-api/src/resolvers/mutation.ts
```diff
@@ -421,3 +490,70 @@ export async function exportList(
...
+  context: IContext,
+): Promise<
+  { url: string; ttl: number } | { message: string; refreshInHours: number }
+> {
+  // Once per day, use the date as key
+  // Hard-coded to a zipfile for now (omnivore) - TODO
+  const filename = `${new Date().toISOString().split('T')[0]}.zip`;
+  const fileKey = path.join(context.userId, importType, filename);
+
+  const s3: NodeJsClient<S3Client> = new S3Client({
```
bassrock, line 511: NodeJsClient - this is some new syntactic sugar 👁️

bassrock (approved):
🔥 Damn this is amazing

servers/list-api/src/background/batchImportHandler.ts
kschelonka: yep, I could see us putting the base class in a package since they all extend from it and just implement the process-specific handling. or using another library yeah.

servers/list-api/src/resolvers/mutation.ts
kschelonka: yeah, it's because otherwise the types aren't narrowed for s3 (so you don't get the outputs properly, the streams aren't node streams, etc.)

PR #911, opened 2024-11-01
fix(files): rejecting file and other invalid url types

bassrock:
# Goal

Reject items that are files, ftp, tel links as a bad user input error

https://pocket.sentry.io/issues/5612403011/?project=5761820&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=24h&stream_index=3

### DIFF
--- a/servers/list-api/src/resolvers/mutation.ts
+++ b/servers/list-api/src/resolvers/mutation.ts
@@ -15,7 +15,12 @@ import { ParserCaller } from '../externalCaller/parserCaller';
 import { SavedItemDataService } from '../dataService';
 import * as Sentry from '@sentry/node';
 import { EventType } from '../businessEvents';
-import { getSavedItemTagsMap, atLeastOneOf, ensureHttpPrefix } from './utils';
+import {
+  getSavedItemTagsMap,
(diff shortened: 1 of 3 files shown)


servers/list-api/src/resolvers/mutation.ts
```diff
@@ -49,10 +54,12 @@ export async function upsertSavedItem(
 ): Promise<SavedItem> {
   const savedItemUpsertInput: SavedItemUpsertInput = args.input;
   const savedItemDataService = new SavedItemDataService(context);
+  const url = ensureHttpPrefix(savedItemUpsertInput.url);
 
+  if (!isHttpUrl(url)) {
```
bassrock, line 59: Pulling it out so that its a user input error and not a server error

kschelonka:
The only problem is that this won't let urls without http work -- e.g. getpocket.com alone or www.getpocket.com

bassrock:
@kschelonka its run after we auto append https to urls like that

kschelonka (approved):

PR #920, opened 2024-11-05
feat(eventBridge): moving shareable lists server to event bridge package

bassrock:
# Goal

Move the shareable lists server to the Pocket event bridge package.

Note: I will move the consumers to the package outside of this PR because a new pattern needs to be stood up for the SharedSnowplowConsumer.


packages/event-bridge/src/events/index.ts
```diff
@@ -32,19 +45,9 @@ export type PocketEvent =
...
-  | ArchiveItem
-  | UnarchiveItem
-  | AddTags
-  | ReplaceTags
-  | ClearTags
-  | RemoveTags
-  | RenameTag
-  | DeleteTag
-  | UpdateTitle;
+  | ListEvent
```
bassrock, line 48: I realized moving each "root type" to a Union type, makes it easier to use and be specific in other services.

servers/shareable-lists-api/src/events/utils.ts
```diff
@@ -0,0 +1,40 @@
+import {
```
bassrock, line 1: Symmetrical Enum helpers.  Helps us convert from Prisma enums to the EventBridge package enums. If this is used elsewhere, will turn into a package.

kschelonka (approved):

PR #921, opened 2024-11-05
fix(import): fix empty query if import had no tags

kschelonka:
Fixes an issue where the transaction would be rolled back if no tags were imported, due to an empty query error


servers/list-api/src/resolvers/mutation.ts
```diff
@@ -132,6 +132,14 @@ export async function batchImport(
     for await (const record of args.input) {
       // Copied from upsertSavedItem
       const url = ensureHttpPrefix(record.url);
+      if (!isHttpUrl(url)) {
```
kschelonka, line 135: copied this change in

bassrock (approved):

PR #923, opened 2024-11-05
feat(sqsEventBridge): creating a SQS poller class

bassrock:
# Goal

Switch Shared Snowplow Consumer to use the Event Bridge types

## Details

* Added in a common SQS Queue Poller for Pocket Events in the eventbridge package
* Updated types where necessary to match types in the snowplow consumer
* Made the Snowplow consumer rely on infra for the DLQ

## Out of Scope

* Setting up the shared snowplow consumer to not create or use HTTP based infrastructure that is not needed
* Updating snowplow


packages/event-bridge/src/consumer.ts
```diff
@@ -0,0 +1,53 @@
...
+      handleMessage: async (message) => {
+        return await options.handlePocketEvent(
+          PocketEventBridgeSQSConsumer.handleMessage(message),
+        );
+      },
+    });
+  }
+
+  private static handleMessage(message: Message): IncomingPocketEvent {
+    const pocketEvent = sqsPollerEventBridgeEvent(message);
```
kschelonka, line 34: Can you please try/catch this so we can handle issues with JSON.parse underlying methods (e.g. JSON.parse on an undefined) -- in case we receive messages that don't have expected boilerplate attributes from flowing through SNS

kschelonka: It will fail regardless in this case -- but just so we have more informative errors and better grouping

kschelonka (approved):

PR #927, opened 2024-11-06
feat(eventBridge): moving corpus indexing to use event bridge types

bassrock:
# Goal

Move all corpus & collection events into the event bridge package

Note: Some collection events seem to have empty objects, so this also introduces a cleaner method to remove those empty object types which are invalid.


lambdas/user-list-search-corpus-indexing/src/index.ts
```diff
@@ -34,13 +38,16 @@ import { removeApprovedItem } from './commands/RemoveItem';
  */
 export async function processor(event: SQSEvent): Promise<SQSBatchResponse> {
   const validPayloads: Array<EventPayload> = event.Records.map((record) => {
-    const message = JSON.parse(JSON.parse(record.body).Message);
+    const event = sqsEventBridgeEvent(record);
+    if (event == null || !validDetailTypes.includes(event['detail-type'])) {
```
kschelonka, line 42: if this can be null, and it really shouldn't be... then we should update the `processor` function signature

lambdas/user-list-search-corpus-indexing/src/index.ts
```diff
@@ -34,13 +38,16 @@ import { removeApprovedItem } from './commands/RemoveItem';
  */
 export async function processor(event: SQSEvent): Promise<SQSBatchResponse> {
   const validPayloads: Array<EventPayload> = event.Records.map((record) => {
-    const message = JSON.parse(JSON.parse(record.body).Message);
+    const event = sqsEventBridgeEvent(record);
```
kschelonka, line 41: nit: change this to not be the same name as the outer scope `processor(event: SQSEvent)`

lambdas/user-list-search-corpus-parser-hydration/src/index.ts
```diff
@@ -29,15 +30,18 @@ export async function processor(event: SQSEvent): Promise<SQSBatchResponse> {
 
   const validPayloads: Array<ValidLangEventPayload> = event.Records.map(
     (record) => {
-      const message = JSON.parse(JSON.parse(record.body).Message);
+      const event = sqsEventBridgeEvent(record);
```
kschelonka, line 33: same nit here, reusing `event` const name in a block scope

kschelonka (approved):

packages/event-bridge/src/jsonUtils.ts
```diff
@@ -0,0 +1,36 @@
+import { isArray, isEmpty, isObject, transform } from 'lodash';
+
+/**
+ * For a given object, remove all empty objects from it
+ * @param obj Object to remove empty objects from
+ * @returns Object that has all empty objects removed {}
+ */
+export const removeEmptyObjects = (obj: Record<string, any>): any => {
```
kschelonka, line 8: ugh...

PR #934, opened 2024-11-07
fix(marticle): updates a use ccase where ul nodes could contain spans and be within a div

bassrock:
# Goal

Fix a long time sentry undefined issue https://pocket.sentry.io/issues/4818338635/?project=5591912&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=14d&stream_index=4


servers/parser-graphql-wrapper/src/marticle/marticleParser.spec.ts
```diff
@@ -449,6 +448,82 @@ describe('MarticleParser', () => {
...
+          {
+            level: 0,
+            content: 'Milk',
+          },
+        ],
+      },
+    ];
+    expect(res).toEqual(expected);
+  });
+  it('should minify and parse unorded lists within a div, that has invalid data in the ul', () => {
```
bassrock, line 481: This replicates the Sentry error when ran on the main branch

servers/parser-graphql-wrapper/src/marticle/marticleParser.ts
```diff
@@ -526,6 +559,9 @@ function listTransformer(
   output: Array<MarticleElement | NumberedListElement | ListElement>;
   aggFrom: number | null;
 } {
+  if (!node.hasChildNodes) {
```
bassrock, line 562: This was the fix for the commented on block above, however when adding this, the main marticle output then gets empty arrays added to it.

servers/parser-graphql-wrapper/src/marticle/marticleParser.ts
```diff
@@ -347,9 +348,31 @@ function RootNode(input: string) {
     '<x-marticle id="marticle-root">' + input + '</x-marticle>',
     null,
   );
+  // Remove any blank or whitespace nodes from the body
+  removeWhitespaceNodes(doc.body);
```
bassrock, line 352: We were trimming whitespace in our tests, so instead, this moves it to after we parse the dom.

servers/parser-graphql-wrapper/src/marticle/marticleParser.spec.ts
```diff
@@ -339,6 +339,260 @@ describe('MarticleParser', () => {
...
+    const res = marticleParser.parse(input);
+    const expected = [
+      {
+        __typename: 'MarticleBulletedList',
+        rows: [
+          {
+            level: 0,
+            content: '1-2.',
+          },
+          {
```
kschelonka, line 417: The schema doesn't allow returning unmarseable in lists so we'll have to update the caller function to remove this or pop it out

PR #991, opened 2024-12-10
feat(notes): paginate a user's notes

kschelonka:
Paginate a user's notes:

* Supports sort by UPDATED_AT or CREATED_AT, ascending or descending
* Supports first/after and last/before
* Supports all filters in the schema

The paginator method is generalized, and it could be pulled out into a package if we want to replicate the integration tests I have that run through the apollo server (can just do the function + arbitrary database).

[POCKET-10858]

[POCKET-10858]: https://mozilla-hub.atlassian.net/browse/POCKET-10858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nzeltzer (approved):
Legit impressed

PR #995, opened 2024-12-10
feat(notes): return markdown content

kschelonka:
Return markdown representation of docContent. Does not do validation.

[POCKET-10896]


nzeltzer (approved):
lgtm

PR #1007, opened 2025-01-07
fix(names): namespacing all packages

bassrock:
# Goal

Update all servers, lambdas and infrastructure package names to be namespaced.

This will let us do things like `pnpm run dev --filter=@server/*'


.github/actions/build-lambda/action.yml
```diff
@@ -73,14 +73,4 @@ runs:
         if [[ -z $s3Key ]]; then
             s3Key="${{ github.sha }}.zip"
         fi
-        aws s3 cp ~/${{ github.sha }}.zip s3://${{inputs.s3-bucket}}/${s3Key}
-    - uses: actions/upload-artifact@v4
-      if: inputs.s3-key == ''
-      with:
-        name: ${{inputs.scope}}
```
bassrock, line 80: Artifact names cant have `/` which they now do. We weren't using this artifact anyway so just removing.

kschelonka (approved):

PR #1030, opened 2025-01-17
chore(parser): bump minimum tasks to 12

kschelonka:
The parser-graphql-wrapper is causing latency issues on graphql requests; bump minimum tasks to handle load. Scaling on cpu and memory does not correlate with increasing request latency.


nzeltzer (approved):
lgtm

PR #1033, opened 2025-01-28
fix(v3proxy): add send_guid endpoint

kschelonka:
Routes to v3/send controller. For backwards-compatible Android requests, which are erroring out due to 404 when accessing Home while not logged in. This endpoint is just legacy analytics, so we will ignore the actions in the same way that we ignore them in v3/send (by returning false and an error in the response, but not a failure of the entire request.

[POCkET-10932]


marcin-kozinski (approved):
Thank you!

PR #1038, opened 2025-01-29
fix(deleter): add missing parameter store permissions

kschelonka:
Adds missing permissions to some shared secrets


nzeltzer (approved):
lgtm

PR #1039, opened 2025-01-31
chore(user-api): update minimum tasks from 2-3

kschelonka:
The auto-scaling seems to be aggressive for the load - on the dashboards we can see kind of whiplash and increase in response times. Bumping the minimum tasks from 2 to 3 to alleviate the load whiplash.


Gio2018 (approved):

nzeltzer (approved):
lgtm

PR #1045, opened 2025-03-12
feat: add additional fallback case for parser failure

kschelonka:
Sometimes the title gets saved as Problem loading page


nzeltzer (approved):
lgtm

PR #1049, opened 2025-05-08
chore(export): Add Sentry alerts for data export failures (HNT-545)

katerinachinnappan:
## Goal

Adding Sentry alerts if data exports are reaching a failing threshold (i.e. above 5% for 10 minutes).

Services for export:
* list
* annotations
* shared lists

`account-data-deleter` manages the state of the overall export job

## Next steps

Create a Sentry Metric Alert for export failures.

Alert created: https://pocket.sentry.io/issues/alerts/rules/details/337137/

## Deployment steps

- [x] Deployed to dev

## References

JIRA ticket:

- https://mozilla-hub.atlassian.net/browse/HNT-545

### DIFF
--- a/servers/account-data-deleter/src/dataService/exportStateService.ts
+++ b/servers/account-data-deleter/src/dataService/exportStateService.ts
@@ -67,6 +67,7 @@ export class ExportStateService {
           },
         });
         Sentry.captureException(error);
+        throw error;
       }
     }
   }
--- a/servers/account-data-deleter/src/dataService/exportStateService.ts
+++ b/servers/account-data-deleter/src/dataService/exportStateService.ts
@@ -48,15 +48,26 @@ export class ExportStateService {
       body: result,
     });
     if (result != null && ExportStateService.isComplete(result)) {
-      const signedUrl = await this.getExportUrl(
-        payload.detail.prefix,
-        payload.detail.encodedId,
-      );
-      await this.notifyUser(
-        payload.detail.encodedId,
-        payload.detail.requestId,
-        signedUrl,
-      );
+      try {
+        const signedUrl = await this.getExportUrl(
+          payload.detail.prefix,
+          payload.detail.encodedId,
+        );
+        await this.notifyUser(
+          payload.detail.encodedId,
+          payload.detail.requestId,
+          signedUrl,
+        );
+      } catch(error) {
+        Sentry.addBreadcrumb({
+          message: 'ExportStateService - processUpdate failed',
+          data: {
+            requestId: payload.detail.requestId,
+            encodedId: payload.detail.encodedId,
+          },
+        });
+        Sentry.captureException(error);
+      }
     }
   }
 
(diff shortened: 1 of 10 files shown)


servers/account-data-deleter/src/dataService/exportStateService.ts
```diff
@@ -48,15 +48,26 @@ export class ExportStateService {
...
+        );
+      } catch(error) {
+        Sentry.addBreadcrumb({
+          message: 'ExportStateService - processUpdate failed',
+          data: {
+            requestId: payload.detail.requestId,
+            encodedId: payload.detail.encodedId,
+          },
+        });
+        Sentry.captureException(error);
```
mmiermans, line 69: Do we want to re-throw the error here as well?
```suggestion
        Sentry.captureException(error);
        throw error;
```

mmiermans (commented):
Wow! It looks to me like you did a very thorough job.

servers/v3-proxy-api/src/generated/graphql/types.ts
```diff
@@ -820,6 +811,18 @@ export enum ExpireUserWebSessionReason {
   PasswordChanged = 'PASSWORD_CHANGED'
 }
 
+export type ExportAcknowledgment = {
```
katerinachinnappan, line 814: got updated automatically when running `pnpm build`

servers/account-data-deleter/src/dataService/exportStateService.ts
katerinachinnappan: good catch!

servers/account-data-deleter/src/dataService/exportStateService.ts
```diff
@@ -176,42 +216,55 @@ export class ExportStateService {
...
-          ':ea': Math.floor(now.getTime() / 1000) + 60 * 60 * 24 * 17, // 17 days (longer than parts retention),
-          ':ca': now.toISOString(),
+          ':ss': true,
+          ':sca': payload.detail.timestamp,
         },
         ReturnValues: 'ALL_NEW' as const,
       };
       const result = await this.dynamo.send(new UpdateCommand(input));
       return result.Attributes as ExportRequestRecord;
+    } catch(error) {
```
jpetto, line 260: i think it would be better to create two `try/catch` blocks here - one for the creation of a new export request record, and one for the updating of the status of that record. it would help us get a better targeted error message.

servers/account-data-deleter/src/queueHandlers/exportStateHandler.ts
```diff
@@ -48,13 +49,20 @@ export class ExportStateHandler extends QueuePoller<{ Message: string }> {
...
       if ('Message' in message && message.Message != null) {
         body = JSON.parse(message.Message);
       } else {
-        throw new Error('Invalid message body');
+        const error = new Error('Invalid SQS message body');
+        Sentry.addBreadcrumb({
+          message: 'Invalid SQS message body',
+          data: { messageBody: message },
+        });
+        Sentry.captureException(error);
```
jpetto, line 64: is there a reason you wouldn't want to capture the exception in the `catch` block below instead?

related - i'm not sure how the breadcrumb in the `catch` block will show up if the exception is already captured here...

jpetto (changes requested):
we should clean up the double exception capturing at least - don't want to inflate our exception counts!

servers/account-data-deleter/src/dataService/exportStateService.ts
```diff
@@ -65,17 +77,33 @@ export class ExportStateService {
...
+        this.zipFileKey(encodedId),
+      );
+      if (!zipResponse) {
+        const error = new Error('ExportStateService - getExportUrl - Failed to zip export files');
+        Sentry.addBreadcrumb({
+          message: 'ExportStateService - getExportUrl - create zip archive failed',
+          data: { encodedId, prefix },
+        });
+        Sentry.captureException(error);
+        throw error;
```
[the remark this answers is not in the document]
katerinachinnappan: yep, I think we can add a Sentry breadcrumb in the `(!zipResponse)` clause & capture it in the final `try/catch` block

servers/account-data-deleter/src/queueHandlers/exportStateHandler.ts
katerinachinnappan: good catch, this should be captured in the outer try/catch block

mmiermans (approved):
As far as I can tell you incorporated all the feedback. Looks good to me. 👍

mmiermans (approved):

PR #1050, opened 2025-05-19
HNT-605: Scale out list-api and data-deleter

mmiermans:
# Goal

[HNT-605](https://mozilla-hub.atlassian.net/browse/HNT-605) Scale out List-API, Account-Data-Deleter (it’s also in charge of exports).

### DIFF
--- a/infrastructure/account-data-deleter/src/dataDeleterApp.ts
+++ b/infrastructure/account-data-deleter/src/dataDeleterApp.ts
@@ -94,10 +94,14 @@ export class DataDeleterApp extends Construct {
           threshold: 25,
         },
       },
+      taskSize: {
+        cpu: config.isDev ? 512 : 2048,
+        memory: config.isDev ? 2048 : 16384,
+      },
       alb6CharacterPrefix: config.shortName,
       autoscalingConfig: {
-        targetMinCapacity: config.isProd ? 2 : 1,
-        targetMaxCapacity: config.isProd ? 3 : 1,
+        targetMinCapacity: config.isProd ? 10 : 1,
+        targetMaxCapacity: config.isProd ? 100 : 1,
       },
       cdn: false,
       codeDeploy: {
--- a/infrastructure/list-api/src/main.ts
+++ b/infrastructure/list-api/src/main.ts
@@ -412,8 +412,8 @@ class ListAPI extends TerraformStack {
           'arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy',
       },
       autoscalingConfig: {
-        targetMinCapacity: config.environment === 'Prod' ? 6 : 1,
-        targetMaxCapacity: config.environment === 'Prod' ? 20 : 10,
+        targetMinCapacity: config.environment === 'Prod' ? 20 : 1,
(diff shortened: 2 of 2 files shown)


infrastructure/account-data-deleter/src/dataDeleterApp.ts
```diff
@@ -94,10 +94,14 @@ export class DataDeleterApp extends Construct {
           threshold: 25,
         },
       },
+      taskSize: {
+        cpu: config.isDev ? 512 : 2048,
+        memory: config.isDev ? 2048 : 16384,
+      },
```
mmiermans, line 100: AccountDataDeleter appears to have a memory leak, because its memory usage climbs to 100% over the course of a few weeks. This may be problematic if the memory leak scales with the number of requests, instead of only over time. Giving it more memory should at least make restarts less frequent. 🤞

image

infrastructure/account-data-deleter/src/dataDeleterApp.ts
```diff
@@ -94,10 +94,14 @@ export class DataDeleterApp extends Construct {
...
       alb6CharacterPrefix: config.shortName,
       autoscalingConfig: {
-        targetMinCapacity: config.isProd ? 2 : 1,
-        targetMaxCapacity: config.isProd ? 3 : 1,
+        targetMinCapacity: config.isProd ? 10 : 1,
+        targetMaxCapacity: config.isProd ? 100 : 1,
```
mmiermans, line 104: I'm expecting we'll get a significantly higher volume of exports than normal. This service runs at < 2% normally with 2 task of 0.5 vCPU. 10 tasks with 2 vCPU should be able to handle a significant increase in traffic.

infrastructure/list-api/src/main.ts
```diff
@@ -412,8 +412,8 @@ class ListAPI extends TerraformStack {
           'arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy',
       },
       autoscalingConfig: {
-        targetMinCapacity: config.environment === 'Prod' ? 6 : 1,
-        targetMaxCapacity: config.environment === 'Prod' ? 20 : 10,
+        targetMinCapacity: config.environment === 'Prod' ? 20 : 1,
+        targetMaxCapacity: config.environment === 'Prod' ? 100 : 10,
```
mmiermans, line 416: Over the last 6 weeks list-api ran at 6 tasks almost constantly, with < 10% average CPU usage. Scaling it to a minimum of 20 tasks should give us room to handle a sudden spike in traffic.

infrastructure/list-api/src/main.ts
```diff
@@ -173,7 +173,7 @@ class ListAPI extends TerraformStack {
       },
       taskSize: {
         cpu: config.isDev ? 2048 : 4096,
-        memory: config.isDev ? 4096 : 8192,
+        memory: config.isDev ? 4096 : 30720,
```
mmiermans, line 176: list-api also seems to suffer from a memory leak, restarting tasks about once or twice per day. Increasing the memory size should make these restarts less frequent.

image

katerinachinnappan (approved):

PR #1102, opened 2026-05-01
feat(terraform modules): expose logging for ALB and RDS constructs

jpetto:
## Goal

increase ability to log system activity to better monitor and investigate potential access anomalies. 

- exposes connection logs for ALBs (in addition to access logs)
- exposes cloudwatch log export for RDS

[HNT-2546](https://mozilla-hub.atlassian.net/browse/HNT-2546)

[HNT-2546]: https://mozilla-hub.atlassian.net/browse/HNT-2546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/terraform-modules/src/base/ApplicationLoadBalancer.ts
+++ b/packages/terraform-modules/src/base/ApplicationLoadBalancer.ts
@@ -19,11 +19,11 @@ export interface ApplicationLoadBalancerProps extends TerraformMetaArguments {
   internal?: boolean;
   useCloudfrontManagedPrefixList?: boolean;
   /**
-   * Optional config to dump alb logs to a bucket.
+   * Optional config to dump alb access logs to a bucket.
    */
   accessLogs?: {
     /**
-     * Existing bucket to dump alb logs too, one of existingBucket or bucket must be chosen.
+     * Existing bucket to dump alb logs to, one of existingBucket or bucket must be chosen.
      */
     existingBucket?: string;
 
@@ -33,7 +33,27 @@ export interface ApplicationLoadBalancerProps extends TerraformMetaArguments {
     bucket?: string;
 
     /**
-     * Optional bucket path prefix. If not defined will use server-logs/{service-name}/internal-alb/AWSLogs/{awsaccountid}/elasticloadbalancing/
+     * Optional bucket path prefix. If not defined will use server-logs/{service-name}/alb/AWSLogs/{awsaccountid}/elasticloadbalancing/
+     * Be sure to include a trailing /
+     */
+    prefix?: string;
+  };
+  /**
+   * Optional config to dump alb connection logs to a bucket.
+   */
+  connectionLogs?: {
+    /**
+     * Existing bucket to dump alb logs to, one of existingBucket or bucket must be chosen.
+     */
+    existingBucket?: string;
+
+    /**
+     * Bucket to dump alb logs too, one of existingBucket or bucket must be chosen.
+     */
+    bucket?: string;
+
+    /**
+     * Optional bucket path prefix. If not defined will use server-logs/{service-name}/albConnection/AWSLogs/{awsaccountid}/elasticloadbalancing/
      * Be sure to include a trailing /
      */
     prefix?: string;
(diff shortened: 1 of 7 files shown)


packages/terraform-modules/src/base/ApplicationRDSCluster.spec.ts
```diff
@@ -22,12 +22,13 @@ describe('ApplicationRDSCluster', () => {
     expect(synthed).toMatchSnapshot();
   });
 
-  it('renders a RDS cluster with a database URL', () => {
```
jpetto, line 25: this removed test was a dupe of the one above - the git diff here doesn't make it clear that the duplicate test was removed and a new one for the cloudwatch exporting added.

packages/terraform-modules/src/base/ApplicationLoadBalancer.ts
```diff
@@ -130,17 +151,50 @@ export class ApplicationLoadBalancer extends Construct {
...
+        bucket,
+        enabled: true,
+        prefix,
+      };
+    }
+
+    let connectionLogsConfig: alb.AlbConnectionLogs | undefined = undefined;
+
+    if (config.connectionLogs !== undefined) {
+      const defaultPrefix = `server-logs/${config.prefix.toLowerCase()}/albConnection`;
```
mmiermans: `server-logs` is kebab case, and `albConnection` is camel case. Would `alb-connection` be more conventional?

packages/terraform-modules/src/base/ApplicationRDSCluster.ts
```diff
@@ -150,6 +155,7 @@ export class ApplicationRDSCluster extends Construct {
         ignoreChanges: ['master_username', 'master_password', 'engine_version'],
       },
       provider: config.provider,
+      enabledCloudwatchLogsExports: config.enableCloudwatchLogsExports ?? [],
```
mmiermans: If users set `config.rdsConfig.enabledCloudwatchLogsExports`, then this clobbers it. You could drop the new top-level prop or accept both:
```suggestion
      enabledCloudwatchLogsExports: config.enableCloudwatchLogsExports ?? configCopy.enabledCloudwatchLogsExports ?? []
```

packages/terraform-modules/src/pocket/PocketALBApplication.ts
```diff
@@ -87,12 +87,36 @@ export interface PocketALBApplicationProps extends TerraformMetaArguments {
...
+    /**
+     * Optional bucket path prefix. If not defined will use server-logs/{service-name}/internal-alb/AWSLogs/{awsaccountid}/elasticloadbalancing/
+     * Be sure to include a trailing /
+     */
+    prefix?: string;
+  };
+
+  /**
+   * Optional config to dump alb access logs to a bucket.
+
```
mmiermans, line 115: Minor copy-paste error. It should say something like:
```suggestion
   * Optional config to dump alb connection logs to a bucket.
```

packages/terraform-modules/src/pocket/PocketALBApplication.ts
```diff
@@ -87,12 +87,36 @@ export interface PocketALBApplicationProps extends TerraformMetaArguments {
...
+   */
+  connectionLogs?: {
+    /**
+     * Existing bucket to dump alb connection logs to, one of existingBucket or bucket must be chosen.
      * If using this options, this module assumes that the bucket already exists in your AWS account and has IAM setup according
      * to https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#attach-bucket-policy which is account wide.
```
mmiermans: I think this comment should also needs to be updated for connection logging.

packages/terraform-modules/src/base/ApplicationRDSCluster.ts
jpetto: this is a good catch - we don't need the extra config prop i added at all. we can just continue to rely on the sub-prop of `configCopy`.

mmiermans (approved):

